PR #3, opened 2018-04-10
Add dependencies for subworkflows

meganshand:


meganshand:
I have used it for myself, yes. And I made sure it still works without dependencies, but nothing formal.

PR #6, opened 2018-06-06
making cromshell compatible with newer curl

### LINKED ISSUE #5: Compatibility issues with newer version of curl
After some testing, it seems that the curl command for submit silently fails with the message `(26) read function returned funny value` for some recent versions of curl which causes the script to get back an empty workflowID. It turns out that when no workflowOptions file is submitted that removing `-F workflowOptions=@` from the curl command seems to fix the issue. The script should be updated to avoid empty files if there are none specified.

lbergelson:
using a version of curl newer than curl ~7.43.0 would cause cromshell submit to fail silently
this was due to curl failing when a -F argument was given an empty filename
fixes #5

### DIFF
--- a/cromwell
+++ b/cromwell
@@ -40,7 +40,7 @@ instruct()  # {{{
 }  # }}}
 
 function submit() { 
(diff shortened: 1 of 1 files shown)


lbergelson:
It uses bash magic that I have never seen before and never want to again.

PR #10, opened 2018-09-12
Allow override of CROMWELL_URL

TedBrookings:
Altered behavior to retain default CROMWELL_URL but allow overriding
it by setting CROMWELL_URL as env var.

This resolves the FIXME comment previously at this location.


TedBrookings:
Good call, I updated README.md

lbergelson:
@TedBrookings 👍 Thank you!

PR #15, opened 2018-11-02
Major refactoring of the script, notifications added.

### LINKED ISSUE #1: fix erros when no jobs have been submitted previously
ex:

```
cat: /Users/louisb/.cromshell/last.workfow.id: No such file or directory
using workflow-id ==
parse error: Invalid numeric literal at line 1, column 4
```

### LINKED ISSUE #11: execution-status-count gives confusing jq error if workflow doesn't exist
```
cromwell execution-status-count 4a9f6782-948a-47d9-9d04-4108e7bd102f
using workflow-id == 4a9f6782-948a-47d9-9d04-4108e7bd102f
jq: error (at <stdin>:3): null (null) has no keys
```

This should probably be clearer.

### LINKED ISSUE #13: rename cromwell script to cromshell
The name Cromwell conflicts with the actual `cromwell` command line tools.  Would people object to renaming it `cromshell`?  You could always alias it back if you don't also use a local cromwell.

### LINKED ISSUE #14: Error code of `status` command does not reflect job status
It would be nice if it did.

jonn-smith:
Added the `notify` sub-command, which will spin off a background task in
bash and periodically check to see if the specified workflow ID is
complete.  Upon completion will send an email to the specified email
address.

Also:

Renamed script to cromshell.
Fixes #13

Now `status` will return 1 if the given workflow did not succeed.
Fixes #14

Now displays an error for failed execution-status-count.
Fixes #11

Now displays an error if CROMWELL_LAST_WORKFLOW_FILE doesn't exist.
Fixes #1


cromshell
```diff
@@ -0,0 +1,311 @@
...
+	# we assume arg 1 is the WORKFLOW_ID and arg 2 
+	# is the email address:
+	if [[ $# -gt 1 ]] ; then
+		# Get the workflow ID:
+		populateWorkflowId $1
+		shift
+	fi
+	
+	# Make sure the user gave us a good email address:
+	echo "$1" | grep '@' &> /dev/null
```
lbergelson: why not just `grep -q '@'` instead of dumping it to /dev/null?

cromshell
```diff
@@ -0,0 +1,311 @@
...
+			# Get our sub-command:
+			SUB_COMMAND=$1
+			shift
+			# We'll manually handle the rest of the options:
+			break
+			;;
+    *)
+      error "${SCRIPTNAME}: invalid option: $1" 
+      simpleUsage
+      echo "Try \`$SCRIPTNAME --help' for more information."
```
lbergelson: What's this escaped backtick for?

cromshell
```diff
@@ -0,0 +1,311 @@
...
+{
+  rm -f ${TMPFILELIST}
+}
+
+function at_exit()
+{
+  cleanTempVars
+}
+
+function checkPipeStatus() {
```
lbergelson: oh, neat, I didn't know about that.

cromshell
```diff
@@ -0,0 +1,311 @@
...
+}
+
+#Display a message to std error:
+function error() 
+{
+  echo "$1" 1>&2 
+}
+
+TMPFILELIST=''
+function makeTemp()
```
lbergelson: Lets remove the stuff that isn't used in the script...

Seems like we could drop the temp file stuff and the exit trapping since those aren't feature we actually use.

cromshell
```diff
@@ -0,0 +1,311 @@
...
+  simpleUsage
+  echo
+  echo -e "When submitting a workflow, the returned worfklow-id will become the" 
+  echo -e "last used workflow id. Also, you can use -n instead of a workflow-id to "
+  echo -e "use a previous run (status -1 would be the status of the last run," 
+  echo -e "-2 would be the one before that).  However, for commands that "
+  echo -e "accepts multiple workflow-id's (status and abort), this does not apply, "
+  echo -e "the explicit workflow-id's are expected if you provide multiple ones."
+  echo
+	echo -e "NOTE: As a convenience, if you omit a workflow-id from a command,"
```
lbergelson: weird spacing here

cromshell
```diff
@@ -0,0 +1,311 @@
...
+		else 
+			WORKFLOW_ID=$(cat ${CROMWELL_LAST_WORKFLOW_FILE} )
+		fi
+  fi
+}
+
+# Submit a workflow and arguments to the Cromwell Server
+function submit() { 
+  local response=$(curl -s -F workflowSource=@${1}  ${2:+ -F workflowInputs=@${2}} ${3:+ -F workflowOptions=@${3}} ${4:+ -F workflowDependencies=@${4}} ${CROMWELL_URL}/api/workflows/v1)
+	local r=$?
```
lbergelson: I think there's some tab / space confusion going on because some lines are wierdly over-indented

cromshell
```diff
@@ -0,0 +1,311 @@
...
+	echo "$1" | grep '@' &> /dev/null
+	r=$?
+	[ $r -ne 0 ] && error "Error: invalid email address: $1" && exit 5
+	email=$1
+
+	# Make sure our workflow ID is populated:
+	populateWorkflowId ${WORKFLOW_ID}
+	
+	# Spin off our notifier:
+	notify $WORKFLOW_ID $email &
```
lbergelson: This should probably be disowned.

cromshell
```diff
@@ -0,0 +1,311 @@
...
+	return $( checkPipeStatus )
+}
+
+# Get the metadata for a Cromwell job UUID
+function metadata() {
+  curl --compressed -s ${CROMWELL_URL}/api/workflows/v1/${1}/metadata?${CROMWELL_METADATA_PARAMETERS} | jq . ;
+	return $( checkPipeStatus )
+}
+
+# Get the metadata for a Cromwell job UUID
```
lbergelson: Description here isn't right.

cromshell
```diff
@@ -0,0 +1,311 @@
...
+# Get the logs of a Cromwell job UUID
+function logs() {
+  curl -s ${CROMWELL_URL}/api/workflows/v1/${1}/logs | jq . ;
+	return $( checkPipeStatus )
+}
+
+# Get the metadata for a Cromwell job UUID
+function metadata() {
+  curl --compressed -s ${CROMWELL_URL}/api/workflows/v1/${1}/metadata?${CROMWELL_METADATA_PARAMETERS} | jq . ;
+	return $( checkPipeStatus )
```
lbergelson: Would `Set +e` accomplish a similar thing to the pipe status checks?   It seems like it's reporting any non-0 exit code for any commands that were connected to the pipe?

cromshell
```diff
@@ -0,0 +1,311 @@
...
+  echo -e "Supported Subcommands:"
+  echo -e "   submit [wdl-file] [inputs] [options] [dependencies]"
+  echo -e "   status [worfklow-id] [[worfklow-id]...]"
+  echo -e "   logs [worfklow-id] [[worfklow-id]...]"
+  echo -e "   metadata [worfklow-id] [[worfklow-id]...]"
+  echo -e "   slim-metadata [worfklow-id] [[worfklow-id]...]"
+  echo -e "   execution-status-count [worfklow-id] [[worfklow-id]...]"
+  echo -e "   timing [worfklow-id] [[worfklow-id]...]"
+  echo -e "   abort [worfklow-id] [[worfklow-id]...]"
+	echo -e "   notify [workflow-id] email_address"
```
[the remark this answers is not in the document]
jonn-smith: I think this is probably a tab-to-spaces issue like you said below.  I'll clear up the file and fix it all.

cromshell
jonn-smith: This is how I learned to do it.  I'll change it to use `-q`.

Old habits die hard I guess.

cromshell
jonn-smith: To display as a back tick.  It's showing the command to run to see the options for the script in quotes, so it'll be rendered as ``cromshell --help'.

cromshell
jonn-smith: Yeah - it's VERY helpful when checking whether a chain of commands has passed (and really the only way).

cromshell
jonn-smith: I like the idea of leaving things in so that they can be used in other scripts people make.

There's a tendency to reuse scripts as templates for future scripts.

I have a script template, from which I pulled most of the refactoring, that I use almost every time I make a new one.  If we can make a template for people to use (or if you want to use mine) then I'd be fine with it.

cromshell
jonn-smith: Definitely tabs/spaces.

Fixed.

cromshell
jonn-smith: Definitely tabs/spaces.

Fixed.

cromshell
jonn-smith: `nohup` added.

cromshell
jonn-smith: Copy / paste error.

Fixed.

cromshell
jonn-smith: Yes, but then you get the `bash` equivalent of a stack trace.

I'll update the check status method to be a little more descriptive.

jonn-smith:
@lbergelson back to you for another review.

I also addressed several issues while I was cleaning it up.

cromshell
jonn-smith: Also, with the latest refactoring, I made it use the `makeTemp` function after all.

cromshell
```diff
@@ -0,0 +1,457 @@
...
+
+# Spin off a task in the background that will check periodically if the workflow ID is 
+# complete.
+# When it is, send an email to the email address specified.
+function notify() {
+
+  local id=$1
+  local email=$2
+  
+  statusFile=$( mktemp )
```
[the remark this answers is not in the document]
jonn-smith: Now reading into a variable.

Fixed!

cromshell
```diff
@@ -0,0 +1,457 @@
+#!/usr/bin/env bash
+
+################################################################################
+
+# Setup variables for the script:
+UNALIASED_SCRIPT_NAME=$( readlink "${BASH_SOURCE[0]}" || echo "${BASH_SOURCE[0]}" )
+SCRIPTDIR="$( cd "$( dirname "${UNALIASED_SCRIPT_NAME}" )" && pwd )"
```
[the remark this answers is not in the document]

PR #21, opened 2018-12-04
Now keeps track of what cromwell server you run on.

### LINKED ISSUE #16: Keep track of which cromwell server a job was run on
Since people use multiple cromwell servers it would be good if we kept track of which one each job ran on.  That way when you run a command like `status -2` you don't have to manually switch the server environment variable if you're running in multiple different environments.

jonn-smith:
- Now keeps track of what cromwell server you run on
- Moved workflow id folders to $CROMSHELL_CONFIG_DIR
- Changed structure of data stored in $CROMSHELL_CONFIG_DIR
- Updated default cromwell server to v36
- Added the `list` command to show what you've run before

Fixes #16

### DIFF
--- a/cromshell
+++ b/cromshell
@@ -8,7 +8,7 @@ SCRIPTDIR="$( cd "$( dirname "${UNALIASED_SCRIPT_NAME}" )" && pwd )"
 SCRIPTNAME=$( echo $0 | sed 's#.*/##g' )
 MINARGS=1
 MAXARGS=99
-PREREQUISITES="curl jq mail perl"
+PREREQUISITES="curl jq mail"
 
 # Determine if this shell is interactive:
 ISCALLEDBYUSER=true
@@ -254,7 +254,7 @@ function status() {
 
   # Update ${CROMWELL_SUBMISSIONS_FILE}:
   local st=$( cat $f | jq . | grep status | sed -e 's#.*: ##g' | tr -d '",' )
-  perl -p -i -e "s#(.*${1}.*\\.wdl)\\t*.*#\\1\\t${st}#g" ${CROMWELL_SUBMISSIONS_FILE}
+  sed -i .bak -e "s#\\(.*${1}.*\\.wdl\\)\\t*.*#\\1$(printf '\t')${st}#g" ${CROMWELL_SUBMISSIONS_FILE}
 
   return $retVal
 }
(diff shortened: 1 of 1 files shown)


mwalker174:
The v36 server is going to be used for running the SV pipeline on gnomAD v3, possibly exclusively because we may need to wipe its database occasionally. Let me ping some folks and get back to you.

cromshell
```diff
@@ -277,35 +298,67 @@ function execution-status-count() {
...
 }
 
+# List all jobs submitted in ${CROMWELL_SUBMISSIONS_FILE}
+function list() {
+	cat ${CROMWELL_SUBMISSIONS_FILE}
+	return $?
+}
+
+# Get the root log folder from the cloud and put it in the metadata folder for this run
+function fetch() {
```
lbergelson: should we rename this fetch-logs?  It's not clear what it's going to fetch

cromshell
```diff
@@ -277,35 +298,67 @@ function execution-status-count() {
...
+}
+
 # Spin off a task in the background that will check periodically if the workflow ID is 
 # complete.
 # When it is, send an email to the email address specified.
 function notify() {
 
   local id=$1
   local email=$2
+	local cromwellServer=$3
```
lbergelson: accidental tab here

cromshell
```diff
@@ -384,20 +438,27 @@ if ${ISCALLEDBYUSER} ; then
...
-      populateWorkflowId $1
+      populateWorkflowIdAndServerUrl $1
       shift
     fi
  
     # Get the host on which this will be run:
     hostServer="${HOSTNAME}"
-    if [[ $# -gt 1 ]] ; then
+    
+		if [[ ! "${1}" =~ ^.*@.*  ]] ; then
```
lbergelson: is this matching something with an `@` in it?

cromshell
jonn-smith: I was wondering about this.  Sure.

cromshell
jonn-smith: Yeah - I need to go through and de-tab the whole thing. 

I have vim to preserve tabs since I read a lot of TSVs.

cromshell
jonn-smith: Yes.  The full purpose (with the negation) is to detect if the next argument is NOT the email address to which to send the notification.

jonn-smith:
Changed the `fetch-logs` command as it was to only grab log files.
Added `fetch-all` that will grab all cromwell output.

PR #25, opened 2018-12-07
only check for the server when necessary

lbergelson:
* this lets you do cromshell list when offline
* fixed typo


lbergelson:
@jonn-smith good catch, back to you

PR #33, opened 2018-12-10
Fixing python version.

jonn-smith:


cromshell
```diff
@@ -3,7 +3,7 @@
 ################################################################################
 
 # Setup variables for the script:
-UNALIASED_SCRIPT_NAME=$( python -c "import os;print os.path.realpath(\"${BASH_SOURCE[0]}\")" )
+UNALIASED_SCRIPT_NAME=$( /usr/bin/python -c "import os;print os.path.realpath(\"${BASH_SOURCE[0]}\")" )
```
lbergelson:
```suggestion
UNALIASED_SCRIPT_NAME=$( python -c "import os;print( os.path.realpath(\"${BASH_SOURCE[0]}\"))" )
```

jonn-smith: Good idea.  Fixed.

PR #34, opened 2018-12-11
fix ping url

huangzhibo:
When my CROMWELL_URL is "http://localhost:8000",  the ":8000" string should be trimed.

### DIFF
--- a/cromshell
+++ b/cromshell
@@ -194,7 +194,7 @@ function checkPrerequisites
 function assertCanCommunicateWithServer
 {
   # Make sure we can talk to the cromwell server:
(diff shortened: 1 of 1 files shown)


cromshell
```diff
@@ -194,7 +194,7 @@ function checkPrerequisites
 function assertCanCommunicateWithServer
 {
   # Make sure we can talk to the cromwell server:
-  serverName=$( echo ${1}| sed 's#.*://##g' )
+  serverName=$( echo ${1}| sed 's#.*://##g'| sed 's#:.*##g' )
```
jonn-smith: It would be good to make this more restrictive with anchors and numerals only.

Can you make the sed statement `'s#:[0-9]*$##g'`

PR #36, opened 2018-12-14
Expand subworkflow metadata

mwalker174:
Cromwell by default does not return subworkflow metadata, probably to cut down on size (I believe you need to enter the subworkflow id to get it, which is a huge pain if you have many subworkflows). Luckily it can be done with the API if you pass this extra parameter `expandSubWorkflows=true`. This PR enables this behavior. What do you guys think?


lbergelson:
@mwalker174 Thanks!

PR #38, opened 2018-12-14
Print turtle to stderr

mwalker174:
The turtle has invaded metadata output in stdout, and I can't find any place where he really needs to be there. This moves him to stderr where he (read I) will be happier.


lbergelson:
@mwalker174 This seems reasonable.  The turtle is meant to be a friend, not an obstruction to analysis.

PR #40, opened 2018-12-19
add note to readme about bioconda

lbergelson:


lbergelson:
@notestaff  I added a line about bioconda.

PR #44, opened 2019-01-25
Dramatic dead turtle

mwalker174:
I don't know if turtles turn upside down when they die, but this makes it easier to recognize failures.


lbergelson:
@mwalker174 Thank you for this important contribution to cromshell!

PR #45, opened 2019-02-04
Refactored to use getopts in some places.

### LINKED ISSUE #42: Update cromshell to use getopts
We live in the future, so our arg parsing should at least live in the present.

### LINKED ISSUE #43: Sometimes the mascot is shown twice.
Turtles are nice
but not when they're twice.
One is enough,
Two is too rough.

jonn-smith:
Could not remove all other arg parsing due to positional args.

Fixed #43 - Twyseturtles was due to the `notify` logic.
Fixed #42 

@lbergelson - I tested it reasonably well locally, but it could definitely use more testing.  In particular, `submit`, `notify`, and `list` (but really a little of everything).

### DIFF
--- a/cromshell
+++ b/cromshell
@@ -28,6 +28,7 @@ COLOR_RUNNING='\033[0;30;46m'
 COLOR_FAILED='\033[1;37;41m'
 COLOR_SUCCEEDED='\033[1;30;42m'
 COLOR_RUNNING='\033[0;30;46m'
+
 ################################################################################
 
 # read env var, or use default server URL
@@ -53,6 +54,7 @@ fi
 ################################################################################
 
 SUB_COMMAND=''
+WORKFLOW_ID_LIST=''
 WORKFLOW_ID=''
 WORKFLOW_SERVER_URL=''
 OPTIONS=''
@@ -101,20 +103,24 @@ function thoughtBubble()
 
 function logo()
 {
-	echo
-	echo '  +--------------------------+'
-	echo '  |        CROMSHELL         |'
-	echo '  +--------------------------+'
+  echo
+  echo '  +--------------------------+'
+  echo '  |        CROMSHELL         |'
+  echo '  +--------------------------+'
   turtle
-	echo
+  echo
 }
 
 ################################################################################
 
 function simpleUsage()
 {
-  echo -e "Usage:    ${SCRIPTNAME} <subcommand> [options]"
-  echo -e "Run and inspect workflows on a Cromwell server."
+  if [[ $# -ne 0 ]] ; then
+    usage | grep ${1} | sed -e 's#\(.*]\).*#\1#g' -e "s#^[ \\t]*#Usage:    ${SCRIPTNAME} #g"
+  else
+    echo -e "Usage:    ${SCRIPTNAME} SUB-COMMAND [options]"
+    echo -e "Run and inspect workflows on a Cromwell server."
+  fi
 }
 
 #Define a usage function:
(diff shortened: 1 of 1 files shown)


cromshell
```diff
@@ -294,15 +371,24 @@ function populateWorkflowIdAndServerUrl()
...
-      assertHavePreviouslySubmittedAJob
-      WORKFLOW_ID=$( tail -n1 ${CROMWELL_SUBMISSIONS_FILE} | awk '{print $3}' )
+      populateLastWorkflowId
       WORKFLOW_SERVER_URL=$(tail -n1 $CROMWELL_SUBMISSIONS_FILE | awk '{print $2}' )
     fi
   fi
+
+  # Validate our workflow ID:
+  if [[ $( matchWorkflowId ${WORKFLOW_ID} ) -ne 0 ]] ; then
+    if [ -n "$userSpecifiedId" ] ; then
```
lbergelson: seems like a good check to add

cromshell
```diff
@@ -432,19 +518,29 @@ function abort()
...
+        ;;
+      u)
+        doUpdate=true
+        ;;
+      *)
+        invalidSubCommand list flag ${OPTARG}
+        ;; 
+    esac 
+  done
+  shift $((OPTIND-1))
```
lbergelson: I don't really know how getopts works.  What's this shift doing after the opt parsing loop? Should it go within the opt parsing loop to loop over all the options?

lbergelson: Also, where is `opt` defined?

cromshell
```diff
@@ -608,20 +704,86 @@ function fetch-all()
...
+    error "Error: You must specify an email address to notify."
+    error ""
+    simpleUsage notify
+    echo "Try \`$SCRIPTNAME -h' for more information."
+    exit 5
+  fi
 
+  # Is the next argument our workflow ID?
+  # Workflow IDs are standard UUIDs or negative numbers:
+  if [[ $(matchWorkflowId ${1}) -eq 0 ]] || [[ ${1} == -* ]] ; then
```
lbergelson: does `[[ ${1} == -* ]]` read as: " the value of $1 is equal to a `-` followed by any characters" ?  If so should it be a match of `-[:digit:]+` instead?

cromshell
```diff
@@ -608,20 +704,86 @@ function fetch-all()
...
+    shift
+  fi
+
+  # Make sure the user gave us a good email address:
+  echo "$1" | grep -q  '@' 
+  r=$?
+  if [ $r -ne 0 ] ; then
+    error "Error: invalid email address: $1"
+    error ""
+    simpleUsage notify
```
lbergelson: Maybe pull this assert out as a function `assertValidEmailAddress`?

cromshell
```diff
@@ -649,127 +811,66 @@ function runSubCommandOnWorkflowId()
...
-        ;;
-      submit|status|logs|execution-status-count|metadata|slim-metadata|timing|abort|notify|list|fetch-all|fetch-logs|list-outputs) 
-        # Get our sub-command:
-        SUB_COMMAND=$1
-        shift
-        # We'll manually handle the rest of the options:
-        break
+  # Get top-level arguments to script itself:
+  while getopts "h" opt ; do
+    case ${opt} in
```
lbergelson: This help checking seems redundant / strange considering the more advanced help check you added elsewhere.

cromshell
```diff
@@ -649,127 +811,66 @@ function runSubCommandOnWorkflowId()
...
+  shift $((OPTIND-1)) 
 
-  ################################################################################
+  # Get our sub-command:
+  SUB_COMMAND=${1}
+  shift
 
-  # Make sure we have all the required commands:
+  # Validate our sub-command:
+  case ${SUB_COMMAND} in
```
lbergelson: additional help check, shouldn't that all be consolidated where you ran the advanced help check?

cromshell
```diff
@@ -649,127 +811,66 @@ function runSubCommandOnWorkflowId()
...
-    shift
-  done
-
-  # Exit the script with the return value of the last run sub-command:
-  exit $r
-
+  # Handle specific sub-command args and and call our sub-command:
+  echo "Sub-Command: ${SUB_COMMAND}"
+  case ${SUB_COMMAND} in 
+    submit|list|notify)
```
lbergelson: this is a huge improvement over the previous status quo

cromshell
```diff
@@ -182,15 +191,25 @@ function makeTemp()
...
 {
   cleanTempVars
 }
+# Make sure we clean up on exit:
+trap at_exit EXIT 
+
+# Print the name and value of a simple variable:
+function printVar()
+{
+  error "$1 = ${!1}"
```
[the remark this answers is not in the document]
jonn-smith: :D

cromshell
```diff
@@ -243,6 +262,64 @@ function checkPrerequisites
...
+      shift
+    done
+  fi
+}
+
+function matchWorkflowId() 
+{
+  local rv=1
+  
+  if [[ "${1}" =~ ^[0-9a-fA-F]{8}-[0-9a-fA-F]{4}-[0-9a-fA-F]{4}-[0-9a-fA-F]{4}-[0-9a-fA-F]{12}$ ]] ; then
```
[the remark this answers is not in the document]
jonn-smith: I specifically want to validate only on valid `UUID`s, so I need to exclude characters `g-z`.

cromshell
jonn-smith: Thanks.  I thought so too :P

cromshell
jonn-smith: It's just a more sane option parser.  The only problem is that while it works well for single character flags, it doesn't work well for positional args or long args (e.g. `--help`), so it ended up being a little less useful for cromshell as it stands now.

That `shift` is shifting the arguments by the number of args that have been parsed by `getopt`.

`opt` is defined at the top of the while loop (line 526): `while getopts "cu" `**opt**` ; do`

lbergelson: Oh, it's defined by getopts... that makes sense... Does getopt default to using `$@` as it's input?

cromshell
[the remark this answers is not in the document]
jonn-smith: xdigit works!

cromshell
jonn-smith: That's my understanding, which is how it can be used in functions (providing you mark `OPTIND` as local).

PR #46, opened 2019-02-04
Added a `cleanup` command to cromshell.

### LINKED ISSUE #29: Add the ability to clear out results from the list command
Currently the list command prints out all of the submitted tasks in your history. Unfortunately that is unbounded and as time goes on I suspect it will start to print out hundreds of forgotten job statuses that nobody wants anymore. A better solution would be to make the list default to returning the last n commands with an argument to print out more if necessary. An alternative would be to add a cleanup command to the history that removes old jobs. Alternatively still the list could be made scrollable though that sounds like a more serious change.

jonn-smith:
Fixes #29


cromshell
```diff
@@ -321,6 +329,26 @@ function invalidSubCommand()
...
 
+function getAnswerFromUser()
+{
+  local prompt="${1}" 
+  local acceptableValues="${2}"
+  local responseVar="${3}" 
+
+  local haveGoodValue=false
+  while ! $haveGoodValue ; do
+    read -p "${prompt} [$( echo ${acceptableValues} | sed 's# #/#g' )] : " ${responseVar}
```
lbergelson: maybe this should be tr?

cromshell
```diff
@@ -321,6 +329,26 @@ function invalidSubCommand()
...
+
+    for okVal in ${acceptableValues} ; do
```
lbergelson: can we make this case insensitive?  ` | tr a-z A-Z` according to stack overflow

cromshell
```diff
@@ -127,48 +129,56 @@ function simpleUsage()
...
+  echo -e ""
   echo -e "  Display a list jobs submitted through cromshell:"
   echo -e "   list [-c] [-u]                                            "
-  echo -e "         -c    Color the output by completion status"
-  echo -e "         -u    Check completion status of all unfinished jobs"
-  echo -e
+  echo -e "         -c               Color the output by completion status."
+  echo -e "         -u               Check completion status of all unfinished jobs."
+  echo -e "" 
+  echo -e "  Clean up local cached list:"
```
lbergelson: add to the readme as well

cromshell
```diff
@@ -321,6 +329,26 @@ function invalidSubCommand()
...
+  while ! $haveGoodValue ; do
+    read -p "${prompt} [$( echo ${acceptableValues} | sed 's# #/#g' )] : " ${responseVar}
+
+    for okVal in ${acceptableValues} ; do
+      if [[ "${!responseVar}" == "${okVal}" ]] ; then
+        haveGoodValue=true
+      fi
+    done
+
+    ! $haveGoodValue && error "Please enter one of the following (CASE SENSITIVE): $( echo ${acceptableValues} | sed 's# #/#g' )" && error ""
```
lbergelson: lets make this tr for consistency

cromshell
jonn-smith: Sure.

cromshell
jonn-smith: Yup!  Good catch!

cromshell
```diff
@@ -601,6 +629,91 @@ function list()
...
+
+  getAnswerFromUser "Are you sure you want to CLEAR ALL RECORDS from these states: ${st}" 'Yes No' answer
+  
+  if [[ "${answer}" == "Yes" ]] ; then
+    echo "User answered 'Yes'."
+    echo
+
+    #Backup the file:
+    echo "Creating backup of records file:"
+    cp -v ${CROMWELL_SUBMISSIONS_FILE} ${CROMWELL_SUBMISSIONS_FILE}.$(date +%Y%m%dT%H%M%S ).bak
```
lbergelson: very nice

cromshell
```diff
@@ -127,48 +129,56 @@ function simpleUsage()
...
-  echo -e "         -u    Check completion status of all unfinished jobs"
-  echo -e
+  echo -e "         -c               Color the output by completion status."
+  echo -e "         -u               Check completion status of all unfinished jobs."
+  echo -e "" 
+  echo -e "  Clean up local cached list:"
+  echo -e "   cleanup [-s STATUS]    Remove completed jobs from local list."
+  echo -e "                          Will remove all jobs from the local list that are in a completed state,"
+  echo -e "                          where a completed state is one of: $(echo ${TERMINAL_STATES} | tr ' ' ',' )" 
+  echo -e "         -s STATUS        If provided, will only remove jobs with the given STATUS from the local list."
```
lbergelson: should we have a -y flag which lets you skip the prompt and shoot yourself in the foot?

cromshell
jonn-smith: Yup.  Fixed.

cromshell
jonn-smith: Thanks.  This seemed important.

cromshell
jonn-smith: For now, I'd say no.  I don't want to let the user be too reckless for now.

cromshell
lbergelson: This is a funny message because you don't have a chance to re-enter it.  It should either be something like `Error: one of the following was required` or it should loop to let you try again.

jonn-smith: It does loop around.  The `while` loop only exits if it validates the input from the user.

The `for` loop is doing an exact strict match on allowable values.

lbergelson: woops, didn't read carefully enough, sorry

PR #47, opened 2019-02-05
Added in basic first-time run config.

### LINKED ISSUE #17: Cromshell should prompt for a server on the first use
Instead of hardcoding a cromwell server, cromshell should prompt you to enter one the first time you run it and then store that value for later.

jonn-smith:
* Added logic to submit to prevent adding errored jobs to list.
* Reworked how logo is displayed for failures.
* Fixed case sensitivity when asking user for input.
* Now tries to use `dialog` and falls back to basic text input if dialog
is not installed.
* Fixed the notification daemon to work with the server file.

Fixes #17

### DIFF
--- a/cromshell
+++ b/cromshell
@@ -11,8 +11,8 @@ MAXARGS=99
 PREREQUISITES="curl jq mail column"
 
 # Determine if this shell is interactive:
-ISCALLEDBYUSER=true
-[[ "${BASH_SOURCE[0]}" != "${0}" ]] && ISCALLEDBYUSER=false
+ISINTERACTIVESHELL=true
+[[ "${BASH_SOURCE[0]}" != "${0}" ]] && ISINTERACTIVESHELL=false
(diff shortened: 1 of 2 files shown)


cromshell
```diff
@@ -936,6 +1175,22 @@ if ${ISCALLEDBYUSER} ; then
   # Get our sub-command:
   SUB_COMMAND=${1}
   shift
+  
+  # Check if we need to set this up.
+  # Note: because of how `notify` works, we can't require the setup for the notify action.
+  if ${CROMWELL_NEEDS_SETUP} && [[ "${SUB_COMMAND}" != "notify" ]] ; then
```
lbergelson: maybe try without this condition and see if it works?

jonn-smith: Just tested it.  It doesn't work because you're effectively calling it when you run the `ssh` command.

PR #50, opened 2019-02-07
Added in -w flag for submit command.

### LINKED ISSUE #49: Submit action should allow waiting for task to change from `submitted` to some final/running state.
We need to add a `-w` flag that will wait for the task submitted by the user to go from the `Submitted` state to some other state.

jonn-smith:
-w flag will wait for status to change from `Submitted` to anything
else before the script exits.

The flag causes some interactive messages on the terminal.

Fixes #49

### DIFF
--- a/cromshell
+++ b/cromshell
@@ -508,7 +508,7 @@ function waitOnSubmittedStatus()
     if $gotNewStatus ; then
       isDone=true
  
-      # Create the flag file to end the display process:
+      # Delete the flag file to end the display process:
       rm -f ${flagFile}
 
       # Wait for the _turtleSpinner to exit:
@@ -525,7 +525,7 @@ function waitOnSubmittedStatus()
     elif [[ $nChecks -ge 10 ]] ; then
       isDone=true
       
-      # Create the flag file to end the display process:
+      # Delete the flag file to end the display process:
       rm -f ${flagFile}
       
       # Wait for the _turtleSpinner to exit:
--- a/cromshell
+++ b/cromshell
@@ -8,7 +8,7 @@ SCRIPTDIR="$( cd "$( dirname "${UNALIASED_SCRIPT_NAME}" )" && pwd )"
 SCRIPTNAME=$( echo $0 | sed 's#.*/##g' )
 MINARGS=1
 MAXARGS=99
-PREREQUISITES="curl jq mail column"
+PREREQUISITES="curl jq mail column tput rev"
 
 # Determine if this shell is interactive:
 ISINTERACTIVESHELL=true
@@ -52,6 +52,7 @@ FOLDER_URL=$( echo ${CROMWELL_URL} | sed -e 's#ht.*://##g' )
 CROMWELL_METADATA_PARAMETERS="excludeKey=submittedFiles&expandSubWorkflows=true"
 
 CURL_CONNECT_TIMEOUT=5
+CURL_MAX_TIMEOUT=10
 
 PING_CMD='ping -c1 -W1 -w10'
 if [[ $( uname ) == "Darwin" ]] ; then
@@ -100,7 +101,7 @@ function turtleDead()
 function simpleUsage()
 {
   if [[ $# -ne 0 ]] ; then
-    usage | grep ${1} | sed -e 's#\(.*]\).*#\1#g' -e "s#^[ \\t]*#Usage:    ${SCRIPTNAME} #g"
+    usage | grep "  ${1} " | sed -e 's#\(.*]\).*#\1#g' -e "s#^[ \\t]*#Usage:    ${SCRIPTNAME} #g"
   else
     echo -e "Usage:    ${SCRIPTNAME} SUB-COMMAND [options]"
     echo -e "Run and inspect workflows on a Cromwell server."
@@ -125,7 +126,9 @@ function usage()
   echo -e "Supported Subcommands:"
   echo -e ""
   echo -e "  Start/Stop workflows:"
-  echo -e "   submit <wdl> <inputs_json> [options_json] [included_wdl_zip_file] Submit a new workflow"
+  echo -e "   submit [-w] <wdl> <inputs_json> [options_json] [included_wdl_zip_file] Submit a new workflow"
+  echo -e "        -w                     Wait for workflow to transition from 'Submitted' to some other status"
+  echo -e "                               before ${SCRIPTNAME} exits"
   echo -e "        included_wdl_zip_file  Zip file containing any WDL files included in the input WDL"
   echo -e ""
   echo -e "   abort [workflow-id] [[workflow-id]...]                            Abort a running workflow."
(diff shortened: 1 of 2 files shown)


cromshell
```diff
@@ -207,10 +210,28 @@ function cleanTempVars()
   rm -f ${TMPFILELIST}
 }
 
+SUBPROCESSLIST=''
+function trackSubProcess() 
+{
+  SUBPROCESSLIST="${SUBPROCESSLIST} $1"
+}
+function unTrackSubProcess()
```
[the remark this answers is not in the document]
jonn-smith: Removed for simplicity.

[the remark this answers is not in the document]
jonn-smith: Err.... all the subprocess tracking is removed...

cromshell
```diff
@@ -118,7 +119,7 @@ function logo()
 function simpleUsage()
 {
   if [[ $# -ne 0 ]] ; then
-    usage | grep ${1} | sed -e 's#\(.*]\).*#\1#g' -e "s#^[ \\t]*#Usage:    ${SCRIPTNAME} #g"
+    usage | grep " ${1} " | sed -e 's#\(.*]\).*#\1#g' -e "s#^[ \\t]*#Usage:    ${SCRIPTNAME} #g"
```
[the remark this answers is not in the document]
jonn-smith: Yup

cromshell
```diff
@@ -207,10 +210,28 @@ function cleanTempVars()
...
+function unTrackSubProcess() 
+{
+  SUBPROCESSLIST=$( echo ${SUBPROCESSLIST} | tr ' ' '\n' | grep -v '^ *$' | grep -v "^${1}$" | xargs echo )
+}
+
+
+# Function to stop all child processes:
+function stopSubProcesses()
+{
+  kill ${SUBPROCESSLIST}
```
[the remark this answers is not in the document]
jonn-smith: OBE

cromshell
```diff
@@ -416,13 +437,159 @@ function populateWorkflowIdAndServerUrl()
...
+  turtle &> ${tmpTurtFile}
+  sed 's#^  ##g' ${tmpTurtFile} > ${turtFile}
+  rev ${turtFile} | sed -e 's$/$#$g' -e 's#\\#/#g' -e 's$#$\\$g' -e 's#(#)#g' -e 's#)#(#g' > ${turtRFile}
+
+  echo "Waiting on status to progress from 'Submitted' ..."
+
+  # Use tput to save the cursor position:
+  tput sc
+
+  # The turtle starts 7 rows above where the cursor is at this point,
```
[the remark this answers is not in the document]
jonn-smith: 8!

cromshell
```diff
@@ -416,13 +437,159 @@ function populateWorkflowIdAndServerUrl()
...
+  tput cuu 8
+
+  # Set up some loop variables:
+  local i=3
+  local incrementString="+1"
+  local curTurtFile=${turtFile}
+
+  # Reserve the name for a file whose existence will cause 
+  # the following loop to exit.
+  local flagFile="__Z_flagFile_$( date +%S ).tmp"
```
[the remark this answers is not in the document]
jonn-smith: Yup.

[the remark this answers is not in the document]

cromshell
```diff
@@ -416,13 +437,159 @@ function populateWorkflowIdAndServerUrl()
...
+      curTurtFile=${turtRFile}
+    elif [ $i -eq 0 ] ; then
+      incrementString="+1" 
+      curTurtFile=${turtFile}
+    fi
+
+  done & # The & here spins this off as a background thread/process
+
+  # Get the pid of the bg display process:
+  local displayPid=$!
```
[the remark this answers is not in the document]
jonn-smith: Yup.  Deleted!

jonn-smith:
@lbergelson Review changes addressed and rebased on master.  Can you take another look?

cromshell
```diff
@@ -402,12 +405,178 @@ function populateWorkflowIdAndServerUrl()
...
+  rev ${turtFile} | sed -e 's$/$#$g' -e 's#\\#/#g' -e 's$#$\\$g' -e 's#(#)#g' -e 's#)#(#g' > ${turtRFile}
+
+  # Set up some loop variables:
+  local i=3
+  local incrementString="+1"
+  local curTurtFile=${turtFile}
+
+  # Let's loop while the ${flagFile} exists.
+  # Note that this will be spun off into its own thread:
+  while [ -f ${flagFile} ] ; do
```
lbergelson: Why -f instead of -e?  Also, when do we need `[[`` vs `[`?

cromshell
```diff
@@ -402,12 +405,178 @@ function populateWorkflowIdAndServerUrl()
...
+  # This keeps it safe.
+  if [[ $# -ne 1 ]] ; then
+    return 1
+  fi
+
+  local flagFile=$1
+
+  # Let's be fancy and make turtles that are walking while we wait:
+  local tmpTurtFile=$( makeTemp )
+  local turtFile=$( makeTemp )
```
lbergelson: will these be cleaned up when this spun off thread dies or when the main one does?  it doesn't matter which it is, I'm just curious how inheriting traps works.

cromshell
```diff
@@ -402,12 +405,178 @@ function populateWorkflowIdAndServerUrl()
...
+    # We could no longer field the status as `Submitted`, so we're done.
+    if [ $? -eq 1 ] && [ $r -ne 2 ] ; then
+      gotNewStatus=true
+    fi
+
+    # Check if our status changed yet:   
+    if $gotNewStatus ; then
+      isDone=true
+ 
+      # Create the flag file to end the display process:
```
lbergelson: Create -> Delete

cromshell
```diff
@@ -402,12 +405,178 @@ function populateWorkflowIdAndServerUrl()
...
+      tput cuu 8
+
+      # Display the status file:
+      cat ${statusFile} | sed 's#^Using ##g'
+  
+    # If we have waited too long, then we quit:
+    elif [[ $nChecks -ge 10 ]] ; then
+      isDone=true
+      
+      # Create the flag file to end the display process:
```
lbergelson: create -> delete

cromshell
```diff
@@ -402,12 +405,178 @@ function populateWorkflowIdAndServerUrl()
...
+      rm -f ${flagFile}
+      
+      # Wait for the _turtleSpinner to exit:
+      wait
```
lbergelson: same question as above

cromshell
```diff
@@ -602,6 +795,13 @@ function list()
         fi
       fi
     done < ${tmpFile}
+
+    # Remove the flag file to kill the spinner and wait for it to finish:
+    rm ${flagFile}
+    wait
```
lbergelson: same question

cromshell
```diff
@@ -402,12 +405,178 @@ function populateWorkflowIdAndServerUrl()
...
+  echo "Waiting on status to progress from 'Submitted' ..."
+  echo
+
+  # Use tput to save the cursor position:
+  tput sc
+
+  # The turtle starts 9 rows above where the cursor is at this point,
+  # so if we need to move it, we have to move the cursor there before
+  # we output anything:
+  tput cuu 9
```
lbergelson: It's correct that this heigh is different than the wait below?

cromshell
jonn-smith: `-f` tests that the input is a regular file and that it exists.  It's just a more specific test.

The single bracket syntax `[...]` is the traditional shell `test` syntax.  It can check a condition and set an error code.

The double bracket syntax `[[...]]` is similar, but is an "extended" test which allows for things like regex matching.  I've used them somewhat interchangeably in the code, but I don't have a good reason for it.  I tend to put more complex tests in double-braces and simple ones in single-braces.

cromshell
jonn-smith: The trap is for the script as a whole, so they'll only be deleted when the main script itself exits.

cromshell
jonn-smith: Good catch!

cromshell
```diff
@@ -402,12 +405,178 @@ function populateWorkflowIdAndServerUrl()
...
+
+    # Check if our status changed yet:   
+    if $gotNewStatus ; then
+      isDone=true
+ 
+      # Create the flag file to end the display process:
+      rm -f ${flagFile}
+
+      # Wait for the _turtleSpinner to exit:
+      wait
```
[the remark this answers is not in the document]
jonn-smith: `wait` without an argument will wait for all sub-processes to finish before unblocking.

Documentation for osx:
```
wait: wait [n]
    Wait for the specified process and report its termination status.  If
    N is not given, all currently active child processes are waited for,
    and the return code is zero.  N may be a process ID or a job
    specification; if a job spec is given, all processes in the job's
    pipeline are waited for.
```

Documentation for Linux:
```
wait: wait [id]
    Wait for job completion and return exit status.
    
    Waits for the process identified by ID, which may be a process ID or a
    job specification, and reports its termination status.  If ID is not
    given, waits for all currently active child processes, and the return
    status is zero.  If ID is a a job specification, waits for all processes
    in the job's pipeline.
    
    Exit Status:
    Returns the status of ID; fails if ID is invalid or an invalid option is
    given.
```

There is no race condition - the only subprocess at this point is the `_turtleSpinner`.  We wait here while the `_turtleSpinner` loops back and learns that the `flagFile` is gone (because we just removed it).

cromshell
jonn-smith: Good catch!

cromshell
jonn-smith: No problem!  (see above)

cromshell
jonn-smith: See above

cromshell
jonn-smith: Yes - the height is determined by the number of output lines between just before the turtle status is initially displayed and where we are now.  We output an extra line after the initial turtle status that we do not add below (L464: `echo "Waiting on status to progress from 'Submitted' ..."`).

When we return to the main process, the cursor is at the top of the turtle.  Since I want to overwrite the '{"id" ...}' line and the "Waiting on status" line.

{"id":"dadfc5e7-f366-4859-830d-dc2f5effcb1b","status":"Submitted"}
Waiting on status to progress from 'Submitted' ...

PR #54, opened 2019-02-20
Curl timeouts

### LINKED ISSUE #53: Curl timeouts should be configurable.
The timeouts for `curl` should be configurable as options to `cromshell`.

jonn-smith:
Added in -t flag for CURL timeouts.

Fixes #53

### DIFF
--- a/cromshell
+++ b/cromshell
@@ -52,7 +52,7 @@ FOLDER_URL=$( echo ${CROMWELL_URL} | sed -e 's#ht.*://##g' )
 CROMWELL_METADATA_PARAMETERS="excludeKey=submittedFiles&expandSubWorkflows=true"
 
 CURL_CONNECT_TIMEOUT=5
-CURL_MAX_TIMEOUT=10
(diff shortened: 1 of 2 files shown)


cromshell
```diff
@@ -1372,6 +1378,29 @@ if ${ISINTERACTIVESHELL} ; then
   # Check the remaining arguments for help and display it if we have to: 
   checkForComplexHelpArgument $@
 
+  # Get flags:
+  while getopts ":t:" opt ; do
+    case ${opt} in 
+      t)
+        # Ensure that the value given to -t is numerical:
+        echo ${OPTARG} | grep -q '^[0-9][0-9]*$'
```
lbergelson: Is this the same as `^[0-9]+$`?

jonn-smith: Yes, `grep` doesn't support the `+` operator without also supplying the `-E` operator.

I'll just add `-E`.

lbergelson: Ah, I didn't realize that.  Either way is fine.

PR #57, opened 2019-02-25
red dead turtle

SHuang-Broad:
a bit more alarming/brutal.

screen shot 2019-02-25 at 12 16 43 pm

### DIFF
--- a/cromshell
+++ b/cromshell
@@ -23,11 +23,6 @@ COLOR_UNDERLINED='\033[1;4m'
 COLOR_FAILED='\033[1;37;41m'
 COLOR_SUCCEEDED='\033[1;30;42m'
 COLOR_RUNNING='\033[0;30;46m'
-################################################################################
-
-COLOR_FAILED='\033[1;37;41m'
-COLOR_SUCCEEDED='\033[1;30;42m'
-COLOR_RUNNING='\033[0;30;46m'
 
 ################################################################################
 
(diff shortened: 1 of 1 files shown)


cromshell
```diff
@@ -88,12 +88,14 @@ function turtle()
 
 function turtleDead()
 {
-  error "    ,,    ,,     ,,       "
-  error "    \\‾\\,-/‾/====/‾/,     "
-  error "     \\/=<‾><‾><‾><‾>-,   "
-  error "     / (\\‾\\‾|‾/‾/‾/‾     "
-  error "    \\‾x/ ˙'-;-;-'˙       "
-  error "     ‾‾                  "
+  export RED='\033[0;31m'
```
jonn-smith: Can you remove the `RED` and `NC` variables and replace them with the global colors above?

Specifically, `RED` should be `COLOR_FAILED` and `NC` should be `COLOR_NORM`.

SHuang-Broad: Just updated as you suggested, and if this is what you meant (background color red)

screen shot 2019-02-25 at 12 23 48 pm

screen shot 2019-02-25 at 12 26 03 pm

jonn-smith: Yup - I figured it would be fine to reuse the color that was already there.

jonn-smith:
Nevermind - I didn't see that there was another divider below.  I'll squash and merge!

SHuang-Broad:
Thanks!

PR #59, opened 2019-02-26
Added in protections to kill the turtleSpinner on exit.

### LINKED ISSUE #58: Must trap _turtleSpinner thread on exit.
Otherwise rogue turtles will walk around terminals, wreaking havoc on unsuspecting prompts everywhere.

jonn-smith:
No longer will rogue turtles plague the terminals of our future selves.

Fixes #58

### DIFF
--- a/cromshell
+++ b/cromshell
@@ -193,10 +193,28 @@ function cleanTempVars()
   rm -f ${TMPFILELIST}
 }
 
+# Track the given subprocess in our list so we can kill it when we exit.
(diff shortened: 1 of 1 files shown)


jonn-smith:
I'm not sure.  It should kill it always, but I just had a turtle running across my terminal so something happened. 

I `ctrl+c`'d a bunch of times really quickly so I wonder if I was able to interrupt the exit catch at just the right time (in which case this wouldn't help either)...

PR #62, opened 2019-03-06
Add ability to load a cromshell plugin on startup.

lbergelson:
This allows users to easily customize their cromshell.

@jonn-smith What do you think about this idea?  I wanted an easy way to change the default colors without having to add a whole bunch of configuration code.

### DIFF
--- a/cromshell
+++ b/cromshell
@@ -1388,6 +1388,11 @@ function setupCromshell()
 
 ################################################################################
 
+#Activate user plugin if it exists
+if [[ -f ${CROMSHELL_CONFIG_DIR}/plugin.sh ]]; then
+  source ${CROMSHELL_CONFIG_DIR}/plugin.sh
+fi
+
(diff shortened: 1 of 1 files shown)


jonn-smith:
I guess that's fine.  I have no real objection, other than the user possibly putting something bad in that file that then comes back to haunt them.

lbergelson:
Do you think there's a better way that prevents people from shooting themselves in the foot?

lbergelson:
I'm not sure how to encapsulate it like that.  I.e. prevent setting the server through the plugin.  This is just giving the user the ability to override whatever they want which definitely could be used for stupid.

jonn-smith:
Maybe a shell script isn't the right approach, then.  What about a JSON file (or similar)?

lbergelson:
Closing this.  We're addressing plugins as part of cromshell 2.0.

PR #64, opened 2019-03-15
Fixed issue with list-outputs.

### LINKED ISSUE #63: list-outputs doesn't list all outputs
List outputs currently compares with the local disk to see what files have yet to be downloaded.

If you partially download outputs then cancel, the displayed list will be only the files left to be downloaded.

This is bad.

jonn-smith:
Fixes #63

### DIFF
--- a/cromshell
+++ b/cromshell
@@ -984,13 +984,16 @@ function list-outputs()
 
   error "Output files from job ${cromwellServer}:${id} : "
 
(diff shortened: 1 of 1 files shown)


jamesemery:
On second thought you should maybe get rid of the tmpdir after the fact

jonn-smith:
Fixed!

PR #65, opened 2019-03-27
Increase performance of execution-status-count

EvanTheB:
This function would not work on my local because the job was too large for cromwell to give a timely response. Similar logic could be put in many subcommands probably. 

And reduce nesting of output, I changed the output syntax to reduce unnecessary (i think?) nesting.

### DIFF
--- a/cromshell
+++ b/cromshell
@@ -45,6 +45,7 @@ fi
 export CROMWELL_URL=${CROMWELL_URL:-"${CROMWELL_SERVER_FROM_FILE}"}
 FOLDER_URL=$( echo ${CROMWELL_URL} | sed -e 's#ht.*://##g' ) 
 CROMWELL_METADATA_PARAMETERS="excludeKey=submittedFiles&expandSubWorkflows=true"
+CROMWELL_SLIM_METADATA_PARAMETERS="includeKey=executionStatus&includeKey=backendStatus&expandSubWorkflows=true"
 
 CURL_CONNECT_TIMEOUT=5
 let CURL_MAX_TIMEOUT=2*${CURL_CONNECT_TIMEOUT}
@@ -703,7 +704,7 @@ function slim-metadata()
 {
   assertCanCommunicateWithServer $2
   turtle
-  curl --connect-timeout $CURL_CONNECT_TIMEOUT --max-time $CURL_MAX_TIMEOUT --compressed -s "${2}/api/workflows/v1/$1/metadata?includeKey=executionStatus&includeKey=backendStatus&expandSubWorkflows=true" | jq .
+  curl --connect-timeout $CURL_CONNECT_TIMEOUT --max-time $CURL_MAX_TIMEOUT --compressed -s "${2}/api/workflows/v1/$1/metadata?$CROMWELL_SLIM_METADATA_PARAMETERS" | jq .
   checkPipeStatus "Could not connect to Cromwell server." "Could not parse JSON output from cromwell server."
   return $?
 }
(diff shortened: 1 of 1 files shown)


cromshell
```diff
@@ -711,21 +711,23 @@ function slim-metadata()
...
-  curl --connect-timeout $CURL_CONNECT_TIMEOUT --max-time $CURL_MAX_TIMEOUT --compressed -s ${2}/api/workflows/v1/$1/metadata?${CROMWELL_METADATA_PARAMETERS} > $f
-  [[ $? -ne 0 ]] && error "Could not connect to Cromwell server." && return 9
+  if ! curl --connect-timeout "$CURL_CONNECT_TIMEOUT" --max-time "$CURL_MAX_TIMEOUT" --compressed -s "$2/api/workflows/v1/$1/metadata?includeKey=executionStatus&includeKey=backendStatus&expandSubWorkflows=true" > "$f" ; then
+    error "Could not connect to Cromwell server." && return 9
+  fi
 
   # Make sure the query succeeded:
-  if [[ "$( cat $f | jq '.status' | sed -e 's#^"##g' -e 's#"$##g' )" == 'fail' ]] ; then
-    reason=$( cat $f | jq '.message' | sed -e 's#^"##g' -e 's#"$##g' )
+  if [[ "$( < "$f" jq '.status' | sed -e 's#^"##g' -e 's#"$##g' )" == 'fail' ]] ; then
```
jonn-smith: What is the rationale for changing the call to `cat` to the output re-direction?  Is it for speed?

cromshell
```diff
@@ -711,21 +711,23 @@ function slim-metadata()
 # Get the status of the given job and how many times it was run
 function execution-status-count() 
 {
-  assertCanCommunicateWithServer $2
+  assertCanCommunicateWithServer "$2"
```
jonn-smith: I know it's somewhat bad form, but can you remove the quotes from the URL for the server here (to make it consistent with the rest of the script)?

cromshell
```diff
@@ -711,21 +711,23 @@ function slim-metadata()
...
   turtle
   f=$( makeTemp )
-  curl --connect-timeout $CURL_CONNECT_TIMEOUT --max-time $CURL_MAX_TIMEOUT --compressed -s ${2}/api/workflows/v1/$1/metadata?${CROMWELL_METADATA_PARAMETERS} > $f
-  [[ $? -ne 0 ]] && error "Could not connect to Cromwell server." && return 9
+  if ! curl --connect-timeout "$CURL_CONNECT_TIMEOUT" --max-time "$CURL_MAX_TIMEOUT" --compressed -s "$2/api/workflows/v1/$1/metadata?includeKey=executionStatus&includeKey=backendStatus&expandSubWorkflows=true" > "$f" ; then
```
jonn-smith: Can you revert the syntax to the 2 line check (again for consistency with the rest of the script)?

jonn-smith: Is there a reason you removed the `${CROMWELL_METADATA_PARAMETERS}` in addition to adding in the new values (`includeKey=executionStatus&includeKey=backendStatus&expandSubWorkflows=true`) in the request?

EvanTheB: i just copied from slim-metadata immediately above.

EvanTheB:
Hi Jonn,

The timeout was a timeout from cromwell itself, ie even with curl infinite timeouts, cromwell responds with something like 'the server is taking too long to respond'. I suspect this is something to do with magic async akka framework. 

It was being triggered by a > 10k task job I had running. cromwell and mysql frequently using 100% core on some threads. It is also a very old version of cromwell.

I don't mind reverting all the shell style changes, I have shell-check running in my editor and just fixed all the things it complained about.

cromshell
```diff
@@ -714,7 +714,8 @@ function execution-status-count()
   assertCanCommunicateWithServer $2
   turtle
   f=$( makeTemp )
-  curl --connect-timeout $CURL_CONNECT_TIMEOUT --max-time $CURL_MAX_TIMEOUT --compressed -s ${2}/api/workflows/v1/$1/metadata?${CROMWELL_METADATA_PARAMETERS} > $f
+  
+  curl --connect-timeout $CURL_CONNECT_TIMEOUT --max-time $CURL_MAX_TIMEOUT --compressed -s "$2/api/workflows/v1/$1/metadata?includeKey=executionStatus&includeKey=backendStatus&expandSubWorkflows=true" > $f
```
jonn-smith: Ah.  Now that I understand better, can you pull out the `includeKey=executionStatus&includeKey=backendStatus&expandSubWorkflows=true"` string and make it a new variable called `CROMWELL_SLIM_METADATA_PARAMETERS` defined just below `CROMWELL_METADATA_PARAMETERS`?

jonn-smith: After you do that, can you also update the call in `slim-metadata`?

jonn-smith:
OK  @EvanTheB - it's in!  Thanks for the patch!

PR #67, opened 2019-04-16
Pretty execution-status-count implementation

tlangs:
As I've been hacking away at ExomeJointGenotyping, I scripted a way to get a nice, human-readable summary of a workflow's state. It was implemented in my own cromwell client script since cromshell (at least to my knowledge) doesn't communicate with authenticated cromwell servers. I noticed the cromshell has an `execution-status-count` subcommand, but its output is JSON, which is great for computers, but not necessarily for humans (or turtles). I've added a `-p` flag which prettifies the output of `execution-status-count`, resulting in a much nicer, color-coded, compact readout of the executionStatus of a workflow. The `-x` flag will expand sub-workflows, which gives a longer output, but details tasks in a subworkflow.

Screen Shot 2019-04-16 at 10 12 28 AM

Let me know if you're interested in this feature!

### DIFF
--- a/cromshell
+++ b/cromshell
@@ -1501,7 +1501,7 @@ if ${ISINTERACTIVESHELL} ; then
   checkForComplexHelpArgument $@
 
   # Get flags:
-  while getopts ":t:xp" opt ; do
+  while getopts ":t:" opt ; do
     case ${opt} in 
       t)
         # Ensure that the value given to -t is numerical:
@@ -1571,8 +1571,6 @@ if ${ISINTERACTIVESHELL} ; then
       rv=$?
       ;;
     *)
-      # All other tools just take a list of workflow IDs
-
       # We need to separate workflow IDs from sub-command arguments
       args=($@)
       workflowIds=()
--- a/cromshell
+++ b/cromshell
@@ -53,8 +53,6 @@ CROMWELL_SLIM_METADATA_PARAMETERS="includeKey=executionStatus&includeKey=backend
 
 CURL_CONNECT_TIMEOUT=5
 let CURL_MAX_TIMEOUT=2*${CURL_CONNECT_TIMEOUT}
-PRETTY_EXECUTION_STATUS_EXPAND_SUB_WORKFLOWS=false
-PRETTY_EXECUTION_STATUS_COUNT=false
 
 PING_CMD='ping -c1 -W1 -w10'
 if [[ $( uname ) == "Darwin" ]] ; then
@@ -718,6 +716,25 @@ function slim-metadata()
 # Get the status of the given job and how many times it was run
 function execution-status-count()
 {
+  # Handle Arguments:
+  local OPTIND
+  local doPretty=false
+  local doExpandSubWorkflows=false
+  while getopts "px" opt ; do
+    case ${opt} in
+      p)
+        doPretty=true
+        ;;
+      x)
+        doExpandSubWorkflows=true
+        ;;
+      *)
+        invalidSubCommand execution-status-count flag ${OPTARG}
+        ;;
+    esac
+  done
+  shift $((OPTIND-1))
+
   assertCanCommunicateWithServer "$2"
   turtle
   f=$( makeTemp )
@@ -732,9 +749,9 @@ function execution-status-count()
     return 11
   fi
 
-  if ${PRETTY_EXECUTION_STATUS_COUNT}; then
+  if ${doPretty}; then
    # Make it pretty for the user:
-    pretty-execution-status ${1} ${f}
+   pretty-execution-status ${1} ${f} ${doExpandSubWorkflows}
   else
     # Make it json for a computer:
     # {tasks:{status: count}}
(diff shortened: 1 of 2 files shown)


cromshell
```diff
@@ -1447,28 +1527,28 @@ if ${ISINTERACTIVESHELL} ; then
 
   # Validate our sub-command:
   case ${SUB_COMMAND} in
-    cleanup|submit|status|logs|execution-status-count|metadata|slim-metadata|timing|abort|notify|list|fetch-all|fetch-logs|list-outputs)
-      # This is a good sub-command, so we do not need to do anything. 
+    cleanup|submit|status|logs|execution-status-count|pretty-execution-status|metadata|slim-metadata|timing|abort|notify|list|fetch-all|fetch-logs|list-outputs)
```
jonn-smith: You don't need to add `pretty-execution-status` in here since the desired entry point from the user is still `execution-status-count`.

cromshell
```diff
@@ -1400,36 +1472,44 @@ function setupCromshell()
...
-        echo ${OPTARG} | grep -q -E '^[0-9]+$' 
+        echo ${OPTARG} | grep -q -E '^[0-9]+$'
         rv=$?
         [ $rv -ne 0 ] && invalidSubCommand '' "flag value (${OPTARG})" "-${opt} argument must be an integer."
 
         # Set our timeouts:
         CURL_CONNECT_TIMEOUT=${OPTARG}
         let CURL_MAX_TIMEOUT=2*${CURL_CONNECT_TIMEOUT}
         ;;
+      x)
```
jonn-smith: Since these aren't global options I think they should be moved into the `execution-status-count` function and handled there.

It would work largely the same way, but you can reference `list` for a good example.

cromshell
```diff
@@ -43,12 +47,14 @@ fi
 
 # read env var, or use default server URL
 export CROMWELL_URL=${CROMWELL_URL:-"${CROMWELL_SERVER_FROM_FILE}"}
-FOLDER_URL=$( echo ${CROMWELL_URL} | sed -e 's#ht.*://##g' ) 
+FOLDER_URL=$( echo ${CROMWELL_URL} | sed -e 's#ht.*://##g' )
```
jonn-smith: I like the removal of the extra whitespace, but can you pull it out of this PR and into another?  It confuses the substantive changes here.

jonn-smith: Alternatively I have a syntax cleanup branch open that I can do this in (if you want to just revert the trailing whitespace lines).

cromshell
```diff
@@ -43,12 +47,14 @@ fi
...
 CROMWELL_METADATA_PARAMETERS="excludeKey=submittedFiles&expandSubWorkflows=true"
-CROMWELL_SLIM_METADATA_PARAMETERS="includeKey=executionStatus&includeKey=backendStatus&expandSubWorkflows=true"
+CROMWELL_SLIM_METADATA_PARAMETERS="includeKey=executionStatus&includeKey=backendStatus&includeKey=status&expandSubWorkflows=true"
 
 CURL_CONNECT_TIMEOUT=5
 let CURL_MAX_TIMEOUT=2*${CURL_CONNECT_TIMEOUT}
+PRETTY_EXECUTION_STATUS_EXPAND_SUB_WORKFLOWS=false
```
jonn-smith: Can you encapsulate these flags away in the `execution-status-count` function family?  This is related to another comment as well (pulling out the option new parsing into those functions - see below).

cromshell
tlangs: Yup, that was an oversight on my part that I didn't remove after I added the `-p` option.

cromshell
tlangs: Ugh, this is IntelliJ's fault...

cromshell
tlangs: This is going to be a little more complicated since we want `execution-status-count` to still run on the implicit workflow IDs, but there's not current mechanism in the code to do that AND handle subcommand-specific args.

cromshell
tlangs: So, when I pushed, IntelliJ automatically cleaned up the whitespace in the same commit as my changes, so there's no separate commit for the whitespace cleanup 😞

tlangs:
bump

jonn-smith:
OK - can you extract your meaningful changes into a patch and create another PR for them (or revert the spacing changes here) so the repo can remain clean?  I'm working on another syntax PR that will resolve most of these.

tlangs:
@jonn-smith I've updated the PR to exclude the whitespace changes.

jonn-smith:
Awesome.  I'm pulling it down to run a couple of tests before merging.

cromshell
```diff
@@ -1478,11 +1571,25 @@ if ${ISINTERACTIVESHELL} ; then
       rv=$?
       ;;
     *)
-      # All other tools just take a list of workflow IDs:
-      handleArgs_workflow_ids $@
+      # All other tools just take a list of workflow IDs
```
jonn-smith: Can you remove this comment?

cromshell
```diff
@@ -1408,7 +1501,7 @@ if ${ISINTERACTIVESHELL} ; then
   checkForComplexHelpArgument $@
 
   # Get flags:
-  while getopts ":t:" opt ; do
+  while getopts ":t:xp" opt ; do
```
jonn-smith: Since these two flags only get processed by `execution-status-count` we don't need to check for them here.  

They can then be passed as arguments to the command rather than the high level script:

```
./cromshell execution-status-count -p -x a63db8dd-e34e-4156-923f-7d10a68e4cb8 5ea63275-6287-42b9-8be2-1460368a491a
```

cromshell
```diff
@@ -730,11 +753,79 @@ function execution-status-count()
     return 11
   fi
 
-  # Make it pretty for the user:
-  # {tasks:{status: count}}
-  jq '.calls | map_values(group_by(.executionStatus) | map({(.[0].executionStatus): . | length}) | add)' "$f"
+  if ${doPretty}; then
```
jonn-smith: Can you also update the README with these options for `execution-status-count`?

tlangs: Updated!

PR #69, opened 2019-05-23
Improve turtle shell pattern stability under moving conditions

lbergelson:
Important turtle fix for issue reported by @yfarjoun

### DIFF
--- a/cromshell
+++ b/cromshell
@@ -441,7 +441,7 @@ function _turtleSpinner()
   local turtRFile=$( makeTemp )
   turtle &> ${tmpTurtFile}
   sed 's#^  ##g' ${tmpTurtFile} > ${turtFile}
-  rev ${turtFile} | sed -e 's$/$#$g' -e 's#\\#/#g' -e 's$#$\\$g' -e 's#(#)#g' -e 's#)#(#g' > ${turtRFile}
+  rev ${turtFile} | sed -e 's$/$#$g' -e 's#\\#/#g' -e 's$#$\\$g' -e 's#(#)#g' -e 's#)#(#g' -e 's#>#a#g' -e 's#<#>#g' -e 's#a#<#g' > ${turtRFile}
 
   # Set up some loop variables:
   local i=3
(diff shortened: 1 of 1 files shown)


cromshell
```diff
@@ -441,7 +441,7 @@ function _turtleSpinner()
   local turtRFile=$( makeTemp )
   turtle &> ${tmpTurtFile}
   sed 's#^  ##g' ${tmpTurtFile} > ${turtFile}
-  rev ${turtFile} | sed -e 's$/$#$g' -e 's#\\#/#g' -e 's$#$\\$g' -e 's#(#)#g' -e 's#)#(#g' > ${turtRFile}
+  rev ${turtFile} | sed -e 's$/$#$g' -e 's#\\#/#g' -e 's$#$\\$g' -e 's#(#)#g' -e 's#)#(#g' -e 's#>#a#g' -e 's#<#>#g' -e 's#a#<#g' > ${turtRFile}
```
yfarjoun:
```suggestion
  rev ${turtFile} | sed -e 's#/#SLASH#g' -e 's#\\#/#g' -e 's#SLASH#\\#g' -e 's#(#OPEN_PAREN#g' -e 's#)#(#g' -e 's#OPEN_PAREN#)#g' -e 's#>#GREATER_THAN#g' -e 's#<#>#g' -e 's#GREATER_THAN#<#g' > ${turtRFile}
```

yfarjoun: no need to limit ourselves to single character temporaries....and I think it makes the "code" cleaner.

jonn-smith: The multi-character match and replacements take longer, though.

It's also a little confusing to read the line and see that only half of the `(` characters are replaced with `OPEN_PAREN`.

I think rather than this replacement a comment above the line stating why there are multiple replacement strings (for the placeholders between `<` and `>`, for example) is a good solution.

yfarjoun: I doubt that the time difference maters for this use, and I find it easier to read...but, I'm not a maintainer, so I just opine....

jonn-smith: @lbergelson can you update the logic for paren swapping as well?

lbergelson: I've compromised by using TMP as the standard replacement for everything and adding a comment.

PR #74, opened 2019-05-31
Fixed regression of relative job ID failure.

### LINKED ISSUE #73: Negative numbers access is broken.

jonn-smith:
The way the arguments were being parsed changed with the last PR.  This broke negative (relative) job ID handling. 

I pushed the argument handling for `execution-status-count` into that function (like with other functions that take arguments) and now it all works again.  I also updated the README and help information to better reflect the new options for this argument.

I've added comments to steer things better in the future.

Fixes #73

### DIFF
--- a/cromshell
+++ b/cromshell
@@ -143,7 +143,9 @@ function usage()
   echo -e "   status [workflow-id] [[workflow-id]...]                           Check the status of a workflow."
   echo -e "   metadata [workflow-id] [[workflow-id]...]                         Get the full metadata of a workflow."
   echo -e "   slim-metadata [workflow-id] [[workflow-id]...]                    Get a subset of the metadata from a workflow."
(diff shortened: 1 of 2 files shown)


cromshell
```diff
@@ -313,6 +315,18 @@ function handleArgs_workflow_ids()
...
 
+function matchRelativeWorkflowId() 
+{
+  local rv=1
+
+  if [[ "${1}" =~ ^-[[:xdigit:]]*$ ]] ; then
+    rv=0
+  fi
+
+  echo ${rv}
```
yfarjoun: why echo here?

jonn-smith: It keeps this function as a mirror of the `matchWorkflowId` function which also prints out the return value.  The echo can probably be removed from both.

PR #75, opened 2019-05-31
Added in checks for file existence for `submit` inputs.

### LINKED ISSUE #71: Cromshell should check the existence of files in "submit"
when the wdl, zip, or json files I'm using to submit are incorrect or missing, the error doesn't provide much information other than ` Server reports that job was not properly submitted ` which is true, but unhelpful.

jonn-smith:
Fixes #71

### DIFF
--- a/cromshell
+++ b/cromshell
@@ -601,7 +601,10 @@ function waitOnSubmittedStatus()
 # Submit a workflow and arguments to the Cromwell Server
 function submit() 
 {
-  local doWait=false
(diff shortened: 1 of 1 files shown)


lbergelson:
This looks good to me, I have an alternate proposal in #76 that adds also checks that the required wdl and inputs json are specified and that none of the files are empty.

jonn-smith:
Let's go with #76 since it's more full-featured.

PR #76, opened 2019-05-31
checking existence of files

lbergelson:
Check that file inputs exist and are not empty.

Addresses #75  #8

### DIFF
--- a/cromshell
+++ b/cromshell
@@ -284,39 +284,31 @@ function checkPrerequisites
 
 function assertRequiredFileIsNonEmpty()
 {
-
   local fileName=$1
   local fileDescription=$2
-  if [[ -z ${fileName} ]]
-	then
-		error "Error: required file ${fileDescription} was not specified."
-		exit 6
-	else
-		if [[ -f ${fileName} ]]
-		then
-			if [[ -s ${fileName} ]]
-			then
-				return 0
-			else
-				error "Error: file ${fileName} is an empty file."
-				exit 6
-			fi
-		else
-			error "Error: file ${fileName} does not exist."
-			exit 6;
-		fi
-	fi
+  if [[ -z ${fileName} ]]; then
+    error "ERROR: ${fileDescription} must be specified."
+  exit 6
+  else
+    if [[ -f ${fileName} ]]; then
+      if [[ ! -s ${fileName} ]]; then
+        error "ERROR: ${fileDescription} is empty: ${fileName}"
+        exit 6
+      fi
+    else
+      error "ERROR: ${fileDescription} does not exist: ${fileName}"
+      exit 6;
+    fi
+  fi
 }
 
 function assertOptionalFileIsNonEmpty()
 {
-	local fileName=$1
-	if [[ -z ${fileName} ]]
-	then
-		return 0
-	else
-		assertRequiredFileIsNonEmpty ${fileName}
-	fi
+  local fileName=$1
+  local fileDescription=$2
+  if [[ ! -z ${fileName} ]] ; then
+    assertRequiredFileIsNonEmpty "${fileName}" "${fileDescription}"
+  fi
 }
 
 function checkForComplexHelpArgument()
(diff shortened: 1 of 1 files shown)


lbergelson:
@jonn-smith Slightly different solution, I'm also checking that the files are non empty and that the required ones are specified.

cromshell
```diff
@@ -282,7 +282,44 @@ function checkPrerequisites
   return 0
 }
 
-function checkForComplexHelpArgument() 
+function assertRequiredFileIsNonEmpty()
+{
+
+  local fileName=$1
+  local fileDescription=$2
+  if [[ -z ${fileName} ]]
```
jonn-smith: A couple of style things:

- Can you tabs to spaces this file (where 1 tab = 2 spaces)?  I've been trying to avoid `\t` as much as possible.
- All the if/then clauses should occur on the same line ala `if CONDITION ; then` to keep formatting consistent.

cromshell
```diff
@@ -282,7 +282,44 @@ function checkPrerequisites
...
+		error "Error: required file ${fileDescription} was not specified."
+		exit 6
+	else
+		if [[ -f ${fileName} ]]
+		then
+			if [[ -s ${fileName} ]]
+			then
+				return 0
+			else
+				error "Error: file ${fileName} is an empty file."
```
jonn-smith: I would rewrite as `"ERROR: ${fileDescription} is empty: ${fileName}"`

cromshell
```diff
@@ -282,7 +282,44 @@ function checkPrerequisites
...
+				exit 6
+			fi
+		else
+			error "Error: file ${fileName} does not exist."
```
jonn-smith: Similarly, I would rewrite as `"ERROR: ${fileDescription} file does not exist: ${fileName}"`

cromshell
```diff
@@ -282,7 +282,44 @@ function checkPrerequisites
...
+}
+
+function assertOptionalFileIsNonEmpty()
+{
+	local fileName=$1
+	if [[ -z ${fileName} ]]
+	then
+		return 0
+	else
+		assertRequiredFileIsNonEmpty ${fileName}
```
jonn-smith: If you're going to call into this function you should have this also take a description string so you can pass it along and keep the user output consistent.

cromshell
```diff
@@ -621,7 +658,17 @@ function submit()
 
   assertCanCommunicateWithServer $CROMWELL_URL
 
-  local response=$(curl --connect-timeout $CURL_CONNECT_TIMEOUT --max-time $CURL_MAX_TIMEOUT -s -F workflowSource=@${1}  ${2:+ -F workflowInputs=@${2}} ${3:+ -F workflowOptions=@${3}} ${4:+ -F workflowDependencies=@${4}} ${CROMWELL_URL}/api/workflows/v1)
+  local wdl=${1}
+  local json=${2}
+  local optionsJson=${3}
+  local dependenciesZip=${4}
+
+  assertRequiredFileIsNonEmpty "$wdl" "wdl"
```
jonn-smith: It may seem a little redundant, but I would add a check for the number of arguments given to `submit` right here with an error to the user if it's less than 2, then I'd output the `simpleUsage` for `submit` and exit so the user knows what the syntax for that command is:

```
  if [[ $# -lt 2 ]] ; then
    error "ERROR: submit requires at least a WDL and JSON file."
    error ""
    simpleUsage submit
    exit 21
  fi
```

jonn-smith: For style I'd address all these variables with braces ala `${VARIABLE}`.  

I have a style / whitespace branch open but I'd like to start enforcing conventions.

cromshell
```diff
@@ -282,7 +282,44 @@ function checkPrerequisites
   return 0
 }
 
-function checkForComplexHelpArgument() 
+function assertRequiredFileIsNonEmpty()
+{
+
```
jonn-smith: extra line

cromshell
```diff
@@ -282,7 +282,44 @@ function checkPrerequisites
...
+			error "Error: file ${fileName} does not exist."
+			exit 6;
+		fi
+	fi
+}
+
+function assertOptionalFileIsNonEmpty()
+{
+	local fileName=$1
+	if [[ -z ${fileName} ]]
```
jonn-smith: You can simplify this if statement:

```
  if [[ ! -z ${fileName} ]] ; then
    assertRequiredFileIsNonEmpty ${fileName}
  fi
```

cromshell
```diff
@@ -282,7 +282,44 @@ function checkPrerequisites
...
+  local fileName=$1
+  local fileDescription=$2
+  if [[ -z ${fileName} ]]
+	then
+		error "Error: required file ${fileDescription} was not specified."
+		exit 6
+	else
+		if [[ -f ${fileName} ]]
+		then
+			if [[ -s ${fileName} ]]
```
jonn-smith: You can simplify this if statement:

```
  if [[ ! -s ${fileName} ]] ; then
    error "Error: file ${fileName} is an empty file."
    exit 6
  fi
```

cromshell
```diff
@@ -282,7 +282,44 @@ function checkPrerequisites
...
 
-function checkForComplexHelpArgument() 
+function assertRequiredFileIsNonEmpty()
+{
+
+  local fileName=$1
+  local fileDescription=$2
+  if [[ -z ${fileName} ]]
+	then
+		error "Error: required file ${fileDescription} was not specified."
```
jonn-smith: Can you rewrite the error statement (see other similar comments).

cromshell
lbergelson: ooh, didn't even realize i had tabs there..

lbergelson: fixed that

cromshell
lbergelson: done

cromshell
lbergelson: 👍

cromshell
lbergelson: 👍

lbergelson:
@jonn-smith I think I've addressed all your comments.  Thank you.

PR #77, opened 2019-06-03
tidy up code in cleanup function

yfarjoun:
- fixed the ability to have multiple states (-s) in cleanup
- print out records to be removed before removing them

### DIFF
--- a/cromshell
+++ b/cromshell
@@ -1025,7 +1025,7 @@ function cleanup()
           invalidSubCommand cleanup STATUS ${OPTARG}
         fi
 
-        st=${OPTARG}
+        st=(${st} ${OPTARG})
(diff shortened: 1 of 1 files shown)


cromshell
```diff
@@ -1040,47 +1040,54 @@ function cleanup()
   fi  
 
   # Get all states if we don't specify one:  
-  if [[ ${#st} -eq 0 ]] ; then   
+  if [[ ${#st} -eq 0 ]] ; then
+    echo getting all states
```
jonn-smith: Because you output all states as you iterate through them, you don't need to output this message here.

yfarjoun:
@jonn-smith thanks. I fixed it all.

PR #83, opened 2019-07-17
Now directly queries the cromwell instance to prove it exists.

jonn-smith:
Should fix issues associated with port numbers and server existence as a corollary.

### DIFF
--- a/cromshell
+++ b/cromshell
@@ -404,8 +404,9 @@ function getAnswerFromUser()
 function assertCanCommunicateWithServer
 {
   # Make sure we can talk to the cromwell server:
-  serverName=$( echo ${1}| sed 's#.*://##g'| sed 's#:[0-9]*##g' )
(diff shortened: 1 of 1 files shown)


cromshell
```diff
@@ -404,8 +404,9 @@ function getAnswerFromUser()
 function assertCanCommunicateWithServer
 {
   # Make sure we can talk to the cromwell server:
-  serverName=$( echo ${1}| sed 's#.*://##g'| sed 's#:[0-9]*##g' )
-  $PING_CMD $serverName &> /dev/null
+	local f=$(makeTemp)
```
lbergelson: This seems fine to me.  Definitely better than ping.  Is it necessary to write the temp file or can we just check the http return code?  The version endpoint might be the most lightweight one, although backends is cheap as well.

jonn-smith: (we resolved this in person)

PR #85, opened 2019-07-30
Added in a WILL_FAIL status for workflows that will fail.

### LINKED ISSUE #78: Provide "Failing" status
When one job failed, but another is still running, the status of the workflow is "Running". This can be misleading and hide the fact that a task has failed and that the workflow will likely fail. 

this jq filter
```
[.calls|.[]|map(.executionStatus)|.[-1]]|any(.=="Failed")
```
gets me part of the way there, but I don't know how to modify the `printTaskStatus` script so that it converts a "Running" status to a "Failing" status in that case....

jonn-smith:
Added a status to indicate if an in-progress workflow will fail.

Also updated some colors.

Fixes #78

### DIFF
--- a/cromshell
+++ b/cromshell
@@ -18,11 +18,15 @@ ISINTERACTIVESHELL=true
 shopt -s expand_aliases
 
 ################################################################################
+
 COLOR_NORM='\033[0m'
 COLOR_UNDERLINED='\033[1;4m'
 COLOR_FAILED='\033[1;37;41m'
+COLOR_WILL_FAIL='\033[1;31;47m'
 COLOR_SUCCEEDED='\033[1;30;42m'
 COLOR_RUNNING='\033[0;30;46m'
+COLOR_ABORTED='\033[0;30;43m'
+
 TASK_COLOR_RUNNING='\033[0;34m'
 TASK_COLOR_SUCCEEDED='\033[0;32m'
 TASK_COLOR_FAILING='\033[0;33m'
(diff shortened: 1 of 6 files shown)


cromshell
```diff
@@ -731,18 +735,42 @@ function status()
...
     turtleDead
+  elif [[ "${workflowStatus}" == "Running" ]] ; then
+    # OK, status claims this workflow is running fine, but we need to check to see
+    # if there are any failed sub-processes.
+    # To do this, we use the `execution-status-count` logic with some filtering:
+    local tmpExecutionStatusCount=$( makeTemp )
+    local tmpMetadata=$( makeTemp )
+
+    # Get execution status count and filter the metadata down:
+    curl --connect-timeout "$CURL_CONNECT_TIMEOUT" --max-time "$CURL_MAX_TIMEOUT" --compressed -s "${2}/api/workflows/v1/${1}/metadata?$CROMWELL_SLIM_METADATA_PARAMETERS" | jq '.calls | map_values(group_by(.executionStatus) | map({(.[0].executionStatus): . | length}) | add)' > ${tmpExecutionStatusCount}
```
lbergelson: I think we drop some greps and a few lines by replacing this block with something like:
```
curl ... | jq --exit-status '[ ..|.executionStatus? | values | . == "Fail" ] | any' >  /dev/null
```

That will exit with 0 if any value is `Fail` and 1 if no value is fail.  We can invert it with `. !+ "Fail" | all`

cromshell
```diff
@@ -731,18 +735,42 @@ function status()
   r=$?
   [[ $r -eq 0 ]] && retVal=1
 
+  # Hold our status string here:
+  local workflowStatus=$( cat $f | jq . | grep status | sed -e 's#.*: ##g' | tr -d '",' )
```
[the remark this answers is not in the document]

cromshell
jonn-smith: We can make the check more elegant, but I'm caching the execution-status-count results so I can display them to the user later.

I'll still need to keep that file / call around.

cromshell
```diff
@@ -748,12 +748,15 @@ function status()
     local tmpMetadata=$( makeTemp )
 
     # Get execution status count and filter the metadata down:
-    curl --connect-timeout "$CURL_CONNECT_TIMEOUT" --max-time "$CURL_MAX_TIMEOUT" --compressed -s "${2}/api/workflows/v1/${1}/metadata?$CROMWELL_SLIM_METADATA_PARAMETERS" | jq '.calls | map_values(group_by(.executionStatus) | map({(.[0].executionStatus): . | length}) | add)' > ${tmpExecutionStatusCount}
-    grep -vE '{|}' ${tmpExecutionStatusCount} > ${tmpMetadata}
+    curl --connect-timeout "$CURL_CONNECT_TIMEOUT" --max-time "$CURL_MAX_TIMEOUT" --compressed -s "${2}/api/workflows/v1/${1}/metadata?$CROMWELL_SLIM_METADATA_PARAMETERS" > ${tmpMetadata}
+    cat ${tmpMetadata} | jq '.calls | map_values(group_by(.executionStatus) | map({(.[0].executionStatus): . | length}) | add)' > ${tmpExecutionStatusCount}
```
lbergelson: I hadn't realized that you needed to save the file for another purpose.  I think the grep makes plenty of sense given that you need the file formatted like that already.

jonn-smith: It's fine - I like having the `jq` call in there for it to be more explicit.

PR #96, opened 2019-10-08
Fixed notification daemon for different local/remote cromwell servers.

jonn-smith:


cromshell
```diff
@@ -1386,21 +1386,31 @@ function notify()
...
+	local WORKFLOW_ID=$1
+	local email=$2
+	local WORKFLOW_SERVER_URL=$3
+
+	error "Spinning off notification thread to ${email} for"
+	error "    workflow:             ${WORKFLOW_ID}" 
+	error "    from Cromwell server: ${WORKFLOW_SERVER_URL}"
+	error "..."
+	nohup bash -c "source ${SCRIPTDIR}/${SCRIPTNAME}; _notifyHelper ${WORKFLOW_ID} ${email} ${WORKFLOW_SERVER_URL}" &>/dev/null &
+	echo "Spun off thread on PID $!"
```
lbergelson: Should this also be error?

jonn-smith: Yup.  Good catch!

jonn-smith: err... Ideally both it and a previous message would be errors, but making them errors causes messages to be parsed incorrectly.  

I'll file a separate issue for it (#97)

Issue #105, opened 2019-12-02
Bug Report: Cromshell is ignorant of "RetryableFailure"s in WILL_FAIL status


jamesemery:
I have encountered an issue with cromshell where I am running a job with a scatter (and consequently a subworkflow being spawned) that according to a `cromshell list -uc` command is in the following state: 
```
20191202_141144 <CROMWELL SERVER>  2d987529-7803-48af-9534-4960b85c1ef6                                                                                                                         evaluate.wdl                             WILL_FAIL
```
Upon investigation running `cromshell execution-status-count 2d987529-7803-48af-9534-4960b85c1ef6` returns `"Running": 1` as it is still hacking away at the subworkflow: 5e9f7dc2-972b-40dc-91d7-c4538eae7076. When I run the `cromshell execution-status-count 5e9f7dc2-972b-40dc-91d7-c4538eae7076` command I find that the subworkflow has the following status: 
```
    "Done": 42,
    "RetryableFailure": 14,
    "Running": 8
```
Upon inspection all of the failures in the metadata are attributed to PAPI error code 10, which is a known headache inducing error condition where the job is killed that can be circumvented by adding a `maxRetries` argument to your wdl tasks to force cromwell to rerun certain PAPI error codes. Unfortunately the `WILL_FAIL` status seems to be interpreting these failures that cromwell is able to disambiguate to "RetryableFailure" as a failure state making my `cromshell list` output confusing. Perhaps there is a way to detect "RetryableFailure" and exclude those from the `WILL_FAIL` logic?

lbergelson:
Seems like a bug with grepping Failure...

lbergelson:
Ok, two bugs here:
1.  we seem to grepping the tmpMetadata and not the tmpExcecutionStatus count.
2.  we are not recursively examining subworkflows.
replacing our current magic jq expression with 
```
 jq '..|.calls? | values | map_values(group_by(.executionStatus) | map({(.[0].executionStatus): . | length}) | add)' ${tmpMetadata} > ${tmpExecutionStatusCount}
```
Fixes it, but I can't figure out how to do it in jq 1.3.  `recurse` should let you do something similar to `..` bu I guess I don't understand the recurse syntax very well because it doesn't work for me.

PR #107, opened 2019-12-04
Update jq query to understand subworkflows

lbergelson:
* This fixes a case where we would fail to mark doomed workflows as doomed.
* Partial fix https://github.com/broadinstitute/cromshell/issues/105
* We now require jq 1.5


lbergelson:
@jonn-smith I added the version check using a cool feature of sort that I didn't know about before.

PR #111, opened 2019-12-09
Add zsh completion script

lbergelson:
* Adding a completion script for zsh.
* It's fairly complete and completes all the current arguments in a relatively sane way.
* It attempts to complete workflow aliases/ids based on a hardcoded path to the cromshell data tsv.

### DIFF
--- a/completions/_cromshell
+++ b/completions/_cromshell
@@ -1,0 +1,134 @@
+#compdef _cromshell cromshell
+
+function _cromshell {
+    local line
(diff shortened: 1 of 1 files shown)


jonn-smith:
Cool.  Is there a `bash` completion script coming soon too?

lbergelson:
I haven't worked on a bash one yet.  I figured you would be more the expert on that.

PR #114, opened 2020-02-19
Fix execution-status-count -x -p

tlangs:
Cromwell changed the way it returns metadata with `includeKey`, so some extra query params need to be provided to get the same metadata output as before.


cromshell
```diff
@@ -68,7 +68,7 @@ fi
 export CROMWELL_URL=${CROMWELL_URL:-"${CROMWELL_SERVER_FROM_FILE}"}
 FOLDER_URL=$( echo ${CROMWELL_URL} | sed -e 's#ht.*://##g' ) 
 CROMWELL_METADATA_PARAMETERS="excludeKey=submittedFiles&expandSubWorkflows=true"
-CROMWELL_SLIM_METADATA_PARAMETERS="${CROMWELL_SLIM_METADATA_PARAMETERS:-includeKey=executionStatus&includeKey=backendStatus&includeKey=status&includeKey=callRoot&expandSubWorkflows=true}"
+CROMWELL_SLIM_METADATA_PARAMETERS="${CROMWELL_SLIM_METADATA_PARAMETERS:-includeKey=subWorkflowMetadata&includeKey=subWorkflowId&inlcudeKey=id&includeKey=executionStatus&includeKey=backendStatus&includeKey=status&includeKey=callRoot&expandSubWorkflows=true}"
```
lbergelson: typo in `inlcudeKey`

lbergelson:
@tlangs Will this fix break older versions of cromwell?  Do you know what version changed things?

tlangs:
I don't think this will break older cromwells. It just needs extra `includeKey` params. Older cromwells that don't need those params will just return the same metadata. I have no way to test this though. You can give cromwell bogus keys, and it won't complain.

PR #116, opened 2020-02-24
fix 112

### LINKED ISSUE #112: `execution-status-count` doesn't seem to work with tasks in sub-workflows
It seems to only display results from level 1 tasks.
So if any of the level 1 tasks is actually a sub-workflow, it won't display the results from the tasks in that sub-workflow.

SHuang-Broad:
close #112 

prompted by #114 

@kvg take notice

### DIFF
--- a/cromshell
+++ b/cromshell
@@ -68,7 +68,8 @@ fi
 export CROMWELL_URL=${CROMWELL_URL:-"${CROMWELL_SERVER_FROM_FILE}"}
 FOLDER_URL=$( echo ${CROMWELL_URL} | sed -e 's#ht.*://##g' ) 
 CROMWELL_METADATA_PARAMETERS="excludeKey=submittedFiles&expandSubWorkflows=true"
-CROMWELL_SLIM_METADATA_PARAMETERS="${CROMWELL_SLIM_METADATA_PARAMETERS:-includeKey=subWorkflowMetadata&includeKey=subWorkflowId&includeKey=id&includeKey=executionStatus&includeKey=backendStatus&includeKey=status&includeKey=callRoot&expandSubWorkflows=true}"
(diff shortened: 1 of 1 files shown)


SHuang-Broad:
This way if the user provides custom `CROMWELL_SLIM_METADATA_PARAMETERS` (I do), those two keys to reflect changes in the Cromwell API will still be included.
Otherwise the custom setting will put the user in a confused state.

cromshell
```diff
@@ -68,7 +68,8 @@ fi
 export CROMWELL_URL=${CROMWELL_URL:-"${CROMWELL_SERVER_FROM_FILE}"}
 FOLDER_URL=$( echo ${CROMWELL_URL} | sed -e 's#ht.*://##g' ) 
 CROMWELL_METADATA_PARAMETERS="excludeKey=submittedFiles&expandSubWorkflows=true"
-CROMWELL_SLIM_METADATA_PARAMETERS="${CROMWELL_SLIM_METADATA_PARAMETERS:-includeKey=subWorkflowMetadata&includeKey=subWorkflowId&includeKey=id&includeKey=executionStatus&includeKey=backendStatus&includeKey=status&includeKey=callRoot&expandSubWorkflows=true}"
+CROMWELL_SLIM_METADATA_PARAMETERS="${CROMWELL_SLIM_METADATA_PARAMETERS:-includeKey=id&includeKey=executionStatus&includeKey=backendStatus&includeKey=status&includeKey=callRoot&expandSubWorkflows=true}"
```
lbergelson: @SHuang-Broad Could you add a comment here to explain to future us why this is split into two lines?

SHuang-Broad:
@lbergelson done

lbergelson:
@SHuang-Broad Thank you.

PR #119, opened 2020-02-29
update list function, 

SHuang-Broad:
to handle workflows that executed under now-retired cromwell servers.

The culprit is [this line](https://github.com/broadinstitute/cromshell/blob/d144f0b7a9f1361bbdd0a2477b63defda39f83d8/cromshell#L1138) which acted as a blackhole: if one line in the TSV has a cromwell server that is now retired (or even temporarily unreachable), the blockhole eats the error message, leaving us with no clue that all entries in the TSV below that line is NOT updated.

### DIFF
--- a/cromshell
+++ b/cromshell
@@ -1124,16 +1124,16 @@ function list()
     tmpFile=$( makeTemp )
     cp ${CROMWELL_SUBMISSIONS_FILE} ${tmpFile}
 
-    UNREACHABLE_SERVER=""
-    REACHABLE_SERVER=""
+    local unreachable_servers=""
+    local reachable_servers=""
     while read line ; do
       id=$( echo "$line" | awk '{print $3}' )
       # Only update if the ID field is, in fact, a valid ID:
       if [[ $(matchWorkflowId ${id}) -eq 0 ]] ; then
         srv=$( echo "$line" | awk '{print $2}' )
-        if echo $UNREACHABLE_SERVER | grep -q $srv; then
+        if echo ${unreachable_servers} | grep -q ${srv}; then
           continue # we already know that server is unreachable
-        elif echo $REACHABLE_SERVER | grep -q $srv; then
+        elif echo ${reachable_servers} | grep -q ${srv}; then
           # we know that server should be reachable, practically speaking during this update operation
           runSt=$( echo "$line" | awk '{print $5}' )
 
@@ -1142,7 +1142,7 @@ function list()
             status ${id} ${srv} &> /dev/null
           fi
         elif [[ $(curl -m 1 -s ${srv}) ]]; then
-          REACHABLE_SERVER+=" $srv"
+          reachable_servers+=" ${srv}"
           
           runSt=$( echo "$line" | awk '{print $5}' )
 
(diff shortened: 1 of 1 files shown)


cromshell
```diff
@@ -1124,16 +1124,34 @@ function list()
...
+        elif echo $REACHABLE_SERVER | grep -q $srv; then
+          # we know that server should be reachable, practically speaking during this update operation
+          runSt=$( echo "$line" | awk '{print $5}' )
+
+          # get the status of the run if it has not completed:
+          if [[ "${runSt}" != "Failed" ]] && [[ "${runSt}" != "Aborted" ]] && [[ "${runSt}" != "Succeeded" ]] ; then
+            status ${id} ${srv} &> /dev/null
+          fi
+        elif [[ $(curl -m 1 -s ${srv}) ]]; then
+          REACHABLE_SERVER+=" $srv"
```
lbergelson: We try to always use {} around variables ( with the sometimes exception of $1 $2 etc). We're not totally consistent about it but it's good general practice.
```suggestion
          REACHABLE_SERVER+=" ${srv}"
```

cromshell
```diff
@@ -1145,6 +1163,8 @@ function list()
     # Restore cursor position:
     tput rc
   fi
+  pretty_ugly=$(echo $UNREACHABLE_SERVER | sed 's/ /\n/g')
+  error "The following cromwell servers cannot be reached, hence skipped\n${pretty_ugly}\n"
```
lbergelson:
```suggestion
  error "The following cromwell servers cannot be reached and were skipped:\n${pretty_ugly}\n"
```

cromshell
```diff
@@ -1145,6 +1163,8 @@ function list()
     # Restore cursor position:
     tput rc
   fi
+  pretty_ugly=$(echo $UNREACHABLE_SERVER | sed 's/ /\n/g')
```
lbergelson:
```suggestion
  local pretty_ugly=$(echo $UNREACHABLE_SERVER | sed 's/ /\n/g')
```

lbergelson: This should be in an if block so it doesn't print if the unreachable server list is empty.

cromshell
```diff
@@ -1124,16 +1124,34 @@ function list()
     tmpFile=$( makeTemp )
     cp ${CROMWELL_SUBMISSIONS_FILE} ${tmpFile}
 
-    while read line ; do 
+    UNREACHABLE_SERVER=""
```
lbergelson: make these plural, UNREACHABLE_SERVERS and REACHABLE_SERVERS.  

I would also probably make them local variables and lowercase, since they won't persist across operations anyway.

cromshell
SHuang-Broad: I believe the new commit does my part.

cromshell
SHuang-Broad: done

cromshell
SHuang-Broad: done

SHuang-Broad: done

cromshell
SHuang-Broad: done

SHuang-Broad:
@lbergelson  I just want to raise one point, that even with this fix, the problem is not completely fixed. 
That line will still error out if somewhere in the process of updating, communication with the servers fail. Then `status` will error out and we are still left with a partially updated table. 
To fix this completely will probably require a more resilient `status` function.

SHuang-Broad:
I think it is possible to have a practically (not theoretical) fix for that, but I do notice that the cromwell server we are using has been responding better and better (no longer that many "Cannot communication with server" messages), so it may become less and less relevant.

PR #128, opened 2020-05-19
Excluded the notify daemon from the jq version check.

jonn-smith:
This effectively bandages the issue of not being able to use the
notification daemons on Broad filesystems because IT won't upgrade our
version of JQ.

Any other systems that don't have a modern version of JQ are now capable of running the `notify` daemon as well.

### DIFF
--- a/cromshell
+++ b/cromshell
@@ -1873,10 +1873,6 @@ if ${ISINTERACTIVESHELL} ; then
   r=$?
   [[ ${r} -ne 0 ]] && exit 6
 
-  checkJQVersion
-  r=$?
-  [[ ${r} -ne 0 ]] && exit 6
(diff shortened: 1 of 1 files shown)


cromshell
```diff
@@ -1888,6 +1884,15 @@ if ${ISINTERACTIVESHELL} ; then
       ;;
   esac
 
+  # Don't check JQ version if we're notifying.
+	# Why do this, you might ask?
+	# Because it's easier to hack our code than to get Broad IT to update software on our servers.
+  if [[ "${SUB_COMMAND_FOR_DISPLAY}" != "notify" ]] ; then
```
lbergelson, line 1890: good idea

PR #140, opened 2020-10-05
Add the sub-command "update-server"

sjfleming:
Absolutely feel free to shoot this down...

I was wanting to update my cromwell server in cromshell, and I always forget where the config file is located.  I thought it might be nice to have a sub-command which takes care of this.

I tried to make the smallest possible change in the code, so it might not be the most elegant solution... I've just reached in and set `CROMWELL_NEEDS_SETUP=true` if the new `update-server` sub-command is run, which should trigger the setup code that already exists.  Does that seem like an acceptable strategy?

### DIFF
--- a/README.md
+++ b/README.md
@@ -95,6 +95,9 @@ requires `column`, `curl`, `mail`, and [jq](https://stedolan.github.io/jq/)
        where a completed state is one of: `Succeeded`, `Failed`, `Aborted`
      * *`-s STATUS`*     If provided, will only remove jobs with the given `STATUS` from the local list.
   
+   #### Update cromwell server
+   * `update-server`
+     * Change the cromwell server that new jobs will be submitted to.
     
  ### Features:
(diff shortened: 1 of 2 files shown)


jonn-smith:
Huh.  I think that's a pretty good solution for now.  @lbergelson - what do you think? 

I'll check it out and give it a couple of test drives.

lbergelson:
Seems like a good idea to me.  Assuming that it works it seems like a good way to handle it as well. @jamesemery was just complaining about the same thing to me.  Need to update the readme and the tab completion script as well.  I can handle the tab completion though in a separate pr.

sjfleming:
Okay great.  I forgot about the readme, I will update that in a couple hours

sjfleming:
Okay the readme should be updated now.  @jonn-smith did your test drives succeed?

jonn-smith (approved):
👍

PR #143, opened 2020-12-01
Bs status unit test

bshifaw:
Added unit testing scripts for Status function and CromshellConfig module


bshifaw:
Bad pr, restarting with fresh branch

PR #144, opened 2020-12-01
Addtion of unit scripts for status function and cromshellconfig module

bshifaw:


tests/unit/mock_data/all.workflow.database.tsv
```diff
@@ -0,0 +1,6 @@
+DATE	CROMWELL_SERVER	RUN_ID	WDL_NAME	STATUS	ALIAS
+20200110_105202	https://cromwell-v47.dsde-methods.broadinstitute.org	a63aa10c-a43e-4ca7-9be9-c2d2aa08b96d	ValidateBam.wdl Failed
```
SHuang-Broad, line 2: I don't see any of the runs being "aliased", so is not representative of real world scenarios.
It would be better if you have aliases for some of the runs, but not all; this facilitates later tests when aliases are allowed.

tests/unit/test_status.py
```diff
@@ -0,0 +1,30 @@
...
+        cromshell_config = cromshellconfig
+        status = command
+        workflow_id = "f373342d-fcfd-4e97-8f6b-cde1ff6ba3ce"
+        # status.main(cromshell_config, workflow_id)
+        # return_value = status.main.ret_val
+        print(cromshell_config.metadata_parameters)
+
+        # assert cromshell_config.metadata_parameters == 1, "Return value should be 1"
+
+    def test_trivial(self):
```
SHuang-Broad, line 19: what is this used for?

tests/unit/test_status.py
```diff
@@ -0,0 +1,30 @@
...
+
+    def test_get_metadata_status_summary(self):
+        cromshell_config = cromshellconfig
+        status = command
+        workflow_id = "f373342d-fcfd-4e97-8f6b-cde1ff6ba3ce"
+        # status.main(cromshell_config, workflow_id)
+        # return_value = status.main.ret_val
+        print(cromshell_config.metadata_parameters)
+
+        # assert cromshell_config.metadata_parameters == 1, "Return value should be 1"
```
SHuang-Broad, line 17: I don't understand this test

tests/unit/test_cromshellconfig.py
```diff
@@ -0,0 +1,97 @@
...
+import pytest
+import os
+
+
+class TestCromshellConfig:
+    """Test the cromshell config functions and variables"""
+
+    def test_override_slim_metadata_parameters(self):
+        reload(cromshellconfig)
+        default_slim_parameters = (
```
SHuang-Broad, line 13: Avoid redefining default values in tests. Instead read them from the place where they are originally defined (in this case, `cromshellconfig`). This avoids headaches of having to manually sync the two definitions.

tests/unit/test_cromshellconfig.py
```diff
@@ -0,0 +1,97 @@
...
+            "=includeKey=id&includeKey=executionStatus&includeKey=backendStatus"
+            "&includeKey=status&includeKey=callRoot&expandSubWorkflows=true&includeKey"
+            "=subWorkflowMetadata&includeKey=subWorkflowId"
+        )
+        assert cromshellconfig.slim_metadata_parameters == default_slim_parameters, \
```
SHuang-Broad, line 18: Therefore this test may be removed.
In OOP, we may test that the constructors correctly initialize data in instances. In this case there's no ctors involved, so we could forgo that.

tests/unit/test_cromshellconfig.py
```diff
@@ -0,0 +1,97 @@
...
+            "By default slim_metadata_parameters should equal \n" \
+            "{slim_metadata_parameters}"
+
+        test_slim_parameters = (
```
SHuang-Broad, line 22: a less general variable name would be better

tests/unit/test_cromshellconfig.py
```diff
@@ -0,0 +1,97 @@
...
+                "expandSubWorkflows=true"
+                + "&includeKey=subWorkflowMetadata&includeKey=subWorkflowId"
```
SHuang-Broad, line 24: is this line tested?

tests/unit/test_cromshellconfig.py
```diff
@@ -0,0 +1,97 @@
...
+        )
+        cromshellconfig.override_slim_metadata_parameters("expandSubWorkflows=true")
+        assert cromshellconfig.slim_metadata_parameters == test_slim_parameters, \
+            "Slim metadata parameters should be set to the" \
+            "following after being specified \n" \
+            f"{test_slim_parameters}"
+
+    def test_resolve_cromwell_config_server_address(self):
+        reload(cromshellconfig)
+        test_cromwell_url = "https://cromwell-v1.dsde-methods.broadinstitute.org"
```
SHuang-Broad, line 34: let's move this closer to the line where it is actually used, line 39

tests/unit/test_cromshellconfig.py
```diff
@@ -0,0 +1,97 @@
...
+            server_user=test_cromwell_url,
+            workflow_id=mock_workflow_id)
+        assert cromshellconfig.cromwell_server == test_cromwell_url, \
+            f"Cromwell server variable should be set to {test_cromwell_url}"
+
+    def test_config_dir(self):
+        reload(cromshellconfig)
+        assert cromshellconfig.config_dir is not None, "Config variable should be set"
+
+        assert cromshellconfig.config_dir == os.path.join(Path.home(), ".cromshell"), \
```
SHuang-Broad, line 71: Is this stated in the Readme?
If a fixed path is assumed, users should be told about that.
We can move `".cromshell"` to `cromshellconfig.py` and treat it like default slim metadata.

tests/unit/test_cromshellconfig.py
```diff
@@ -0,0 +1,97 @@
...
+        assert Path(cromshellconfig.config_dir).exists(), \
+            "Config directory should exist "
+
+    def test_submission_file(self):
+        reload(cromshellconfig)
+        assert cromshellconfig.submission_file is not None, \
+            "Submission file variable should be set "
+
+        path_to_submission = os.path.join(Path.home(), ".cromshell",
+                                          "all.workflow.database.tsv")
```
SHuang-Broad, line 83: similar to `".cromshell"`, we should move this to `cromshellconfig.py`.

tests/unit/test_cromshellconfig.py
```diff
@@ -0,0 +1,97 @@
...
+        assert cromshellconfig.submission_file == path_to_submission, \
+            f"Submission file path should be {path_to_submission} "
+
+        assert Path(cromshellconfig.submission_file).exists(), \
+            "Submission file should exist"
+
+    def test_cromwell_server(self):
+        reload(cromshellconfig)
+        assert cromshellconfig.cromwell_server is not None, \
```
SHuang-Broad, line 92: There are multiple tests that just tests a file/folder is set (not `None`) and exists, so what about refactoring out an utility method just for that?

SHuang-Broad (changes requested):
@bshifaw Nice start!
Just some comments about general structures.

tests/unit/mock_data/all.workflow.database.tsv
bshifaw: done

PR #145, opened 2020-12-02
explicitly prepend http to server url when IP address is used instead…

SHuang-Broad:
… of an url

Otherwise the `open` command treats the ip address as local file.

### DIFF
--- a/cromshell
+++ b/cromshell
@@ -1088,9 +1088,14 @@ function printTaskStatus()
 # Bring up a browser window to view timing information on the job.
 function timing() 
 {
+  local id=${1}
+  local server_url_for_browser=${2}
   turtle
-  echo "Opening timing information in a web browser for job ID: ${1}"
-  open ${2}/api/workflows/v1/${1}/timing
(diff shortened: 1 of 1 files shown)


cromshell
```diff
@@ -1088,9 +1088,14 @@ function printTaskStatus()
...
 function timing() 
 {
+  local id=${1}
+  local server_url_for_browser=${2}
   turtle
-  echo "Opening timing information in a web browser for job ID: ${1}"
-  open ${2}/api/workflows/v1/${1}/timing
+  echo "Opening timing information in your default web browser for job ID: ${id}"
+  if [[ ${server_url_for_browser} != "http*" ]]; then
+    server_url_for_browser="http://${server_url_for_browser}"
```
lbergelson, line 1096: Seems good to me.  You tested that it works with both IP and still works with addresses?

SHuang-Broad: yes I did: for url I tested some old executions on methods 51, for ip:port I tested with freshly run workflows.

lbergelson (approved):

PR #146, opened 2021-01-15
Fix to be able to iterate through workflow metadata to check subworkflow failures

bshifaw:
Fix to be able to itterate through workflow metadata to obtain subworkflow statuses

### DIFF
--- a/src/cromshell/status/command.py
+++ b/src/cromshell/status/command.py
@@ -126,7 +126,7 @@ def check_for_failure(metadata: dict):
         # and their status within layers of the dictionary. Depth First Search recursion
         # is used to traverse the dictionary by iterating through any dictionary
         # this function comes across first and reporting whether it found a failure
-        if type(value) is dict:
+        if isinstance(value, dict):
             workflow_failed = check_for_failure(value)
 
         # If a list value is encountered then the dictionary being traversed through is
@@ -137,7 +137,7 @@ def check_for_failure(metadata: dict):
         # dictionary holding status for the shard.
         # We'll want to check each shard to determine whether it's status has
         # Failed
-        if type(value) is list:
+        if isinstance(value, list):
             for i, shard in enumerate(value):
                 if "subWorkflowMetadata" in shard.keys():
                     workflow_failed = check_for_failure(shard["subWorkflowMetadata"])
--- a/src/cromshell/status/command.py
+++ b/src/cromshell/status/command.py
@@ -107,27 +107,26 @@ def main(config, workflow_id):
 def check_for_failure(metadata: dict):
     """Checks a workflow metadata dictionary for failing statuses
     Returns True to indicate workflow or some task(s) has failed"""
-    workflow_failed = False
-    # If the given dictionary contains a 'status' key and has
-    # value of "Failed" then exit the function returning
-    # True to indicate workflow has failed
-    if "status" in metadata:
-        if metadata["status"] == "Failed":
-            workflow_failed = True
-            return workflow_failed
-    # If the dictionary does not contain a failed value for
-    # its status key or if status key does does not exist
-    # then we'll iterate through this loop in search for
-    # task status or subworkflows.
-    for key, value in metadata.items():
-        # If a dictionary value is encountered then recursion is used iterate through
-        # the dictionary to find task statuses. The reason for this is due to the tree
-        # structure of the metadata dictionary which holds the calls, subworkflow calls,
-        # and their status within layers of the dictionary. Depth First Search recursion
-        # is used to traverse the dictionary by iterating through any dictionary
-        # this function comes across first and reporting whether it found a failure
+
+    # If the given dictionary contains a 'status' key and has value of "Failed"
+    # then exit the function returning "True" to indicate workflow has failed
+    if metadata.get("status") == "Failed":
+        return True
+
+    # If the dictionary does not contain a failed value for its status key or
+    # if status key does does not exist then we'll iterate through the dictionary
+    # value in search of a nested dictionary (indicating subworkflows)
+    # or a list which holds the task statuses.
+    for value in metadata.values():
+
+        # If a dictionary value is encountered then Depth First Search recursion
+        # is used to traverse the dictionary by iterating through the given dictionary
+        # and reporting whether it found a failure. The reason for this is due to the
+        # tree structure of the metadata dictionary which holds the task statuses
+        # in second layer dictionary key called calls.
         if isinstance(value, dict):
-            workflow_failed = check_for_failure(value)
+            if check_for_failure(value):
+                return True
 
         # If a list value is encountered then the dictionary being traversed through is
         # a key and value pair where the key is the task name and the value is the
@@ -135,17 +134,20 @@ def check_for_failure(metadata: dict):
         # task was scattered. Hence the value for the key is a list to account
         # for all the shards for a task. Each item(shard) within this list is a
         # dictionary holding status for the shard.
-        # We'll want to check each shard to determine whether it's status has
-        # Failed
+        # We'll want to check each shard to determine whether it's status has "Failed".
         if isinstance(value, list):
-            for i, shard in enumerate(value):
+            for shard in value:
+
+                # If a key in the shared labeled "subWorkflowMetadata" then this
+                # will another dictionary layer of subworkflow tasks that needs
+                # to be traversed.
                 if "subWorkflowMetadata" in shard.keys():
-                    workflow_failed = check_for_failure(shard["subWorkflowMetadata"])
+                    if check_for_failure(shard["subWorkflowMetadata"]):
+                        return True
                 else:
                     if shard["executionStatus"] == "Failed":
-                        workflow_failed = True
-                        return workflow_failed
-    return workflow_failed
+                        return True
+    return False
 
 
 if __name__ == "__main__":
--- a/src/cromshell/status/command.py
+++ b/src/cromshell/status/command.py
@@ -57,20 +57,21 @@ def main(config, workflow_id):
         # TODO : Use this as a template for the Metadata subcommand
         # Get execution status count and filter the metadata down:
         request_meta_out = requests.get(
-            f"{config.cromwell_api_workflow_id}/metadata?{config.slim_metadata_parameters} "
+            f"{config.cromwell_api_workflow_id}/metadata?{config.slim_metadata_parameters}"
         )
 
         # tmp_metadata holds the workflow metadata as a dictionary
         tmp_metadata = json.loads(request_meta_out.content.decode("utf-8"))
-        execution_status_count = get_metadata_status_summary(tmp_metadata)
+        summarized_tmp_metadata = get_all_call_values(tmp_metadata)
+        execution_status_count = get_metadata_status_summary(summarized_tmp_metadata)
+        print(execution_status_count)
 
         # Check for failure states:
         failed = False
-        for task_status in execution_status_count:
-            for key in task_status:
-                if "Failed" in str(task_status[key]):
-                    failed = True
-                    break
+        for call in execution_status_count:
+            if "Failed" in str(execution_status_count[call]):
+                failed = True
+                break
 
         # Check for failures:
         if not failed:
@@ -97,14 +98,14 @@ def main(config, workflow_id):
 
     # Update config.submission_file:
     with fileinput.FileInput(
-        config.submission_file, inplace=True, backup=".bak"
+            config.submission_file, inplace=True, backup=".bak"
     ) as csv_file:
         reader = csv.DictReader(csv_file, delimiter="\t")
         print("\t".join(reader.fieldnames))
         for row in reader:
             if (
-                row["CROMWELL_SERVER"] == config.cromwell_server
-                and row["RUN_ID"] == workflow_id
+                    row["CROMWELL_SERVER"] == config.cromwell_server
+                    and row["RUN_ID"] == workflow_id
             ):
                 row["STATUS"] = workflow_status
                 print("\t".join(x for x in row.values() if x))
@@ -114,34 +115,74 @@ def main(config, workflow_id):
     return ret_val
 
 
-def get_metadata_status_summary(workflow_metadata):
+def get_all_call_values(nested_dictionary):
+    """Summarizes a given workflow metadata dictionary into a
+    new dictionary with only task name and status"""
+    temp_dic1 = {}
+    temp_dic2 = {}
+    temp_dic3 = {}
+    for key, value in nested_dictionary.items():
+        # Skip any keys that have values that are strings, because we are note
+        # interested in them
+        if type(value) is str:
+            continue
+        # If a dictionary value is encountered then we iterate through the dictionary
+        # to find call statuses
+        if type(value) is dict:
+            temp_dic1 = get_all_call_values(value)
+        # If a list value is encountered then those are the calls and their status,
+        # we'll want to add the key and value to the temp dictionary
+        if type(value) is list:
+            # Each item in the list is the shard of a call so we'll itterate through
+            # the list and collect info on all shards
+            temp_dic3[key] = []
+            for i, shard in enumerate(value):
+                calls_status = shard
+                # If a key in the shard dictionary has the term "subWorkflowMetadata"
+                # then this is a subworkflow so we iterate through the dictionary to
+                # find call statuses
+                if "subWorkflowMetadata" in calls_status.keys():
+                    temp_dic2 = get_all_call_values(calls_status["subWorkflowMetadata"])
+                    # Remove "Scatter" key from dict
+                    if key in temp_dic3.keys():
+                        temp_dic3.pop(key)
+                # If there isn't a sub workflow then we add the call name combined
+                # with the shard index as a key and its metadata dictionary as its
+                # value
+                if "subWorkflowMetadata" not in calls_status.keys():
+                    temp_dic3[key].append(calls_status)
+
+    # Merge all the dictionaries together from all the different if statments
+    merged_dic = {**temp_dic1, **temp_dic2, **temp_dic3}
+    return merged_dic
+
+
+def get_metadata_status_summary(workflow_call_metadata):
     """Get the status for each call in a workflow
     and the frequency of those statuses"""
-    # workflow_metadata holds the workflow metadata as a dictionary
-    tmp_execution_status = []
 
-    # For each call in the metadata dictionary create a shortened summary containing
-    # the call name and status
-    for call in workflow_metadata["calls"]:
-        call_element = f'"{call}": '
-        execution_statuses = []
+    tmp_execution_status = {}
+
+    # For each call in the given metadata dictionary create a shortened dictionary
+    # containing the call name and status
+    for call in workflow_call_metadata:
+        execution_statuses = []  # List to hold the different status per call
+        tmp_execution_status[call] = {}
 
         # For each call name add the number of execution_status to a list
-        for instanceDic in workflow_metadata["calls"][call]:
-            execution_statuses.append(instanceDic["executionStatus"])
+        for shard in workflow_call_metadata[call]:
+            execution_statuses.append(shard["executionStatus"])
 
         # Create a key and value pair for the execution status and number of times
-        # it is seen
+        # it is seen for this call
         status_count = Counter(execution_statuses)
+
         # For each status for the call add the call name and status name and count
         # to list
         for status in set(execution_statuses):
-            status_element = f'{{"{status}": {status_count[status]}}}'
-            tmp_execution_status.append(f"{call_element}{status_element}")
+            tmp_execution_status[call][status] = status_count[status]
 
-    tmp_execution_status_json = ", ".join(tmp_execution_status)
-    tmp_execution_status_json = "{" + tmp_execution_status_json + "}"
-    return [json.loads(tmp_execution_status_json)]
+    return tmp_execution_status
 
 
 if __name__ == "__main__":
--- a/src/cromshell/status/command.py
+++ b/src/cromshell/status/command.py
@@ -46,7 +46,8 @@ def main(config, workflow_id):
     workflow_status = workflow_status_description["status"]
 
     # Set return value based on workflow status
-    if workflow_status in ("Failed", "Aborted", "fail"):
+    if workflow_status in cromshellconfig.WorkflowStatuses.Failed.value or \
+            cromshellconfig.WorkflowStatuses.Aborted.value:
         ret_val = 1
         log.display_logo(io_utils.dead_turtle)
     elif workflow_status == "Running":
@@ -69,7 +70,7 @@ def main(config, workflow_id):
             log.display_logo(io_utils.turtle)
         else:
             log.display_logo(io_utils.doomed_logo)
-            workflow_status = "DOOMED"
+            workflow_status = cromshellconfig.WorkflowStatuses.DOOMED.value
             message = (
                 "The workflow is Running but one of the instances "
                 "has failed which will lead to failure."
@@ -125,7 +126,7 @@ def workflow_failed(metadata: dict):
         # tree structure of the metadata dictionary which holds the task statuses
         # in a nested dictionary key called "calls".
         if isinstance(value, dict):
-            if check_for_failure(value):
+            if workflow_failed(value):
                 return True
 
         # If a list value is encountered then the dictionary being traversed through is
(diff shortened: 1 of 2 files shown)


src/cromshell/status/command.py
```diff
@@ -114,34 +115,74 @@ def main(config, workflow_id):
...
 
-def get_metadata_status_summary(workflow_metadata):
+def get_all_call_values(nested_dictionary):
+    """Summarizes a given workflow metadata dictionary into a
+    new dictionary with only task name and status"""
+    temp_dic1 = {}
+    temp_dic2 = {}
+    temp_dic3 = {}
+    for key, value in nested_dictionary.items():
+        # Skip any keys that have values that are strings, because we are note
```
SHuang-Broad, line 125: "note" -> "not"

src/cromshell/status/command.py
```diff
@@ -114,34 +115,74 @@ def main(config, workflow_id):
...
+        # interested in them
+        if type(value) is str:
+            continue
+        # If a dictionary value is encountered then we iterate through the dictionary
+        # to find call statuses
+        if type(value) is dict:
```
SHuang-Broad, line 131: It's worth having some comments here about using recursion for DFS in traversing the metadata tree.

src/cromshell/status/command.py
```diff
@@ -114,34 +115,74 @@ def main(config, workflow_id):
     return ret_val
 
 
-def get_metadata_status_summary(workflow_metadata):
+def get_all_call_values(nested_dictionary):
+    """Summarizes a given workflow metadata dictionary into a
+    new dictionary with only task name and status"""
+    temp_dic1 = {}
```
SHuang-Broad, line 121: `temp_dic1/2/3` could use better names to reflect what their intended use/contents are.

src/cromshell/status/command.py
```diff
@@ -114,34 +115,74 @@ def main(config, workflow_id):
     return ret_val
 
 
-def get_metadata_status_summary(workflow_metadata):
+def get_all_call_values(nested_dictionary):
```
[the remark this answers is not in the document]
SHuang-Broad: Also, `nested_dictionary` isn't as intuitive as it can be.
How about `metadata_json: dict` with python 3 type hints.

src/cromshell/status/command.py
```diff
@@ -114,34 +115,74 @@ def main(config, workflow_id):
...
+    """Summarizes a given workflow metadata dictionary into a
+    new dictionary with only task name and status"""
+    temp_dic1 = {}
+    temp_dic2 = {}
+    temp_dic3 = {}
+    for key, value in nested_dictionary.items():
+        # Skip any keys that have values that are strings, because we are note
+        # interested in them
+        if type(value) is str:
```
SHuang-Broad, line 127: `if isinstance(value, str)` is more intuitive.

src/cromshell/status/command.py
SHuang-Broad, line 131: `if isinstance(value, dict)`

src/cromshell/status/command.py
```diff
@@ -114,34 +115,74 @@ def main(config, workflow_id):
...
+            continue
+        # If a dictionary value is encountered then we iterate through the dictionary
+        # to find call statuses
+        if type(value) is dict:
+            temp_dic1 = get_all_call_values(value)
+        # If a list value is encountered then those are the calls and their status,
+        # we'll want to add the key and value to the temp dictionary
+        if type(value) is list:
```
SHuang-Broad, line 135: `if isinstance(value, list)`

src/cromshell/status/command.py
```diff
@@ -114,34 +115,74 @@ def main(config, workflow_id):
...
+    for key, value in nested_dictionary.items():
+        # Skip any keys that have values that are strings, because we are note
+        # interested in them
+        if type(value) is str:
+            continue
+        # If a dictionary value is encountered then we iterate through the dictionary
+        # to find call statuses
+        if type(value) is dict:
+            temp_dic1 = get_all_call_values(value)
+        # If a list value is encountered then those are the calls and their status,
```
SHuang-Broad, line 133: It's worth giving an explanation here or above, on why type is being used to decide what's in it.
We talked offline so we know what's happening, but this might confuse someone who haven't seen a metadata json structure.

src/cromshell/status/command.py
```diff
@@ -114,34 +115,74 @@ def main(config, workflow_id):
...
+        # we'll want to add the key and value to the temp dictionary
+        if type(value) is list:
+            # Each item in the list is the shard of a call so we'll itterate through
+            # the list and collect info on all shards
+            temp_dic3[key] = []
+            for i, shard in enumerate(value):
+                calls_status = shard
```
SHuang-Broad, line 140: I think the variables `shard` and `calls_status` could use better names. `shard` typically indicate there's a scatter-gather problem we are facing, which is not always the case here. `call_status` indicate it's THE status we are looking for, but it's not actually that, itself has rich structures as indicated below.

src/cromshell/status/command.py
```diff
@@ -114,34 +115,74 @@ def main(config, workflow_id):
...
+                # find call statuses
+                if "subWorkflowMetadata" in calls_status.keys():
+                    temp_dic2 = get_all_call_values(calls_status["subWorkflowMetadata"])
+                    # Remove "Scatter" key from dict
+                    if key in temp_dic3.keys():
+                        temp_dic3.pop(key)
+                # If there isn't a sub workflow then we add the call name combined
+                # with the shard index as a key and its metadata dictionary as its
+                # value
+                if "subWorkflowMetadata" not in calls_status.keys():
```
SHuang-Broad, line 152: could be just the `else` block of line 144

src/cromshell/status/command.py
```diff
@@ -114,34 +115,74 @@ def main(config, workflow_id):
...
+            temp_dic3[key] = []
+            for i, shard in enumerate(value):
+                calls_status = shard
+                # If a key in the shard dictionary has the term "subWorkflowMetadata"
+                # then this is a subworkflow so we iterate through the dictionary to
+                # find call statuses
+                if "subWorkflowMetadata" in calls_status.keys():
+                    temp_dic2 = get_all_call_values(calls_status["subWorkflowMetadata"])
+                    # Remove "Scatter" key from dict
+                    if key in temp_dic3.keys():
```
SHuang-Broad, line 147: I don't quite understand the logic here. We can talk "in person" to go over it.

SHuang-Broad (changes requested):
@bshifaw I've done my 1st round. Other than some small change suggestions, I have a question about the logic of pop a dictionary. We can talk "in person" to understand that better.

src/cromshell/status/command.py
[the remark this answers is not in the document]
bshifaw: done

src/cromshell/status/command.py
```diff
@@ -57,20 +57,21 @@ def main(config, workflow_id):
...
-            f"{config.cromwell_api_workflow_id}/metadata?{config.slim_metadata_parameters} "
+            f"{config.cromwell_api_workflow_id}/metadata?{config.slim_metadata_parameters}"
         )
 
         # tmp_metadata holds the workflow metadata as a dictionary
         tmp_metadata = json.loads(request_meta_out.content.decode("utf-8"))
-        execution_status_count = get_metadata_status_summary(tmp_metadata)
+        summarized_tmp_metadata = get_all_call_values(tmp_metadata)
+        execution_status_count = get_metadata_status_summary(summarized_tmp_metadata)
+        print(execution_status_count)
```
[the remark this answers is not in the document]
bshifaw: removed

src/cromshell/status/command.py
bshifaw: done

src/cromshell/status/command.py
bshifaw: done

src/cromshell/status/command.py
bshifaw: done
also did this for the other if statements

src/cromshell/status/command.py
bshifaw: done

src/cromshell/status/command.py
bshifaw: In line 138 dictionary_3 is created using the keys from the dictionary being iterated through (example task1, task2, task3), however these can either be actual tasks or place holders for tasks from subworkflows (example ScatterAt98).  In line 144 we check if any of the associated tasks have subworkflows, if they do then we get their tasks and place into a dictionary_2 then remove the placeholder task from dictionary_3 because its not an actual task but a placeholder for subworkflow tasks.

src/cromshell/status/command.py
bshifaw: done

src/cromshell/status/command.py
bshifaw: done

bshifaw: @SHuang-Broad Added this 
"# If a dictionary value is encountered then recursion is used iterate through
        # the dictionary to find call statuses. The reason for this is due the tree
        # structure of the metadata dictionary which holds the calls, subworkflow calls,
        # and their status within layers of the dictionary. Depth First Search recursion
        # is used to traverse the dictionary by iterating through any dictionary this 
        # this functions comes across first and obtaining their call and status. "

src/cromshell/status/command.py
bshifaw: @SHuang-Broad , Added this:
"# If a list value is encountered then the dictionary being traversed through is
        # a key and value pair where the key is the task name and the value is the
        # number of times the task was executed (shard), which is normally 1 unless the
        # task was scattered. Hence the value for the key is a list to account
        # for all the shards for a task. Each item(shard) within this list is a
        # dictionary holding status for the shard.
        # We'll want to add this key and list value holding all its shard(s)
        # to the temp dictionary."

src/cromshell/status/command.py
bshifaw: @SHuang-Broad Would it be alright to keep the term` shard` and give a definition for it in the comments, like the one that was recently posted ([link](https://github.com/broadinstitute/cromshell/pull/146/files#r561873560)). 
`call_status` was removed, and the for loop now just uses `shard`.

src/cromshell/status/command.py
```diff
@@ -114,34 +104,48 @@ def main(config, workflow_id):
...
-    tmp_execution_status_json = "{" + tmp_execution_status_json + "}"
-    return [json.loads(tmp_execution_status_json)]
+def check_for_failure(metadata: dict):
+    """Checks a workflow metadata dictionary for failing statuses
+    Returns True to indicate workflow or some task(s) has failed"""
+    workflow_failed = False
+    # If the given dictionary contains a 'status' key and has
+    # value of "Failed" then exit the function returning
+    # True to indicate workflow has failed
+    if "status" in metadata:
```
lbergelson, line 114: Instead of nested ifs you could `if use metadata.get("status") == "Failed"`. Maybe not pythonic style though?  I'm not sure.

src/cromshell/status/command.py
```diff
@@ -114,34 +104,48 @@ def main(config, workflow_id):
...
+        # If a list value is encountered then the dictionary being traversed through is
+        # a key and value pair where the key is the task name and the value is the
+        # number of times the task was executed (shard), which is normally 1 unless the
+        # task was scattered. Hence the value for the key is a list to account
+        # for all the shards for a task. Each item(shard) within this list is a
+        # dictionary holding status for the shard.
+        # We'll want to check each shard to determine whether it's status has
+        # Failed
+        if isinstance(value, list):
+            for i, shard in enumerate(value):
```
lbergelson, line 141: Why are you enumerating here if you ignore the number?

src/cromshell/status/command.py
```diff
@@ -114,34 +104,48 @@ def main(config, workflow_id):
...
+                if "subWorkflowMetadata" in shard.keys():
+                    workflow_failed = check_for_failure(shard["subWorkflowMetadata"])
```
lbergelson, line 143: Does this work?  It seems like a failing workflow followed by a successful one in the enumeration will overwrite the return value.  Am I crazy?

src/cromshell/status/command.py
```diff
@@ -114,34 +104,48 @@ def main(config, workflow_id):
...
+    # True to indicate workflow has failed
+    if "status" in metadata:
+        if metadata["status"] == "Failed":
+            workflow_failed = True
+            return workflow_failed
+    # If the dictionary does not contain a failed value for
+    # its status key or if status key does does not exist
+    # then we'll iterate through this loop in search for
+    # task status or subworkflows.
+    for key, value in metadata.items():
```
lbergelson, line 122: It seems like you ignore the key, so why not just traverse the values?

src/cromshell/status/command.py
```diff
@@ -114,34 +104,48 @@ def main(config, workflow_id):
...
+        # If a dictionary value is encountered then recursion is used iterate through
+        # the dictionary to find task statuses. The reason for this is due to the tree
+        # structure of the metadata dictionary which holds the calls, subworkflow calls,
+        # and their status within layers of the dictionary. Depth First Search recursion
+        # is used to traverse the dictionary by iterating through any dictionary
+        # this function comes across first and reporting whether it found a failure
+        if isinstance(value, dict):
+            workflow_failed = check_for_failure(value)
```
lbergelson, line 130: If you end up with true here don't you want to return?

lbergelson (commented):
@bshifaw I think there's a bit of a problem with the recursive conditions returning where a failed case can be overwritten by a subsequent successful one since they don't return immediately.

src/cromshell/status/command.py
```diff
@@ -57,23 +56,14 @@ def main(config, workflow_id):
...
-        failed = False
-        for task_status in execution_status_count:
-            for key in task_status:
-                if "Failed" in str(task_status[key]):
-                    failed = True
-                    break
 
         # Check for failures:
-        if not failed:
+        if not check_for_failure(tmp_metadata):
```
lbergelson, line 66: I would rename this to be more of a statement ` if not workflow_failed` is clearer than `if not check_for_failure`

src/cromshell/status/command.py
bshifaw: Make sense, `use` isn't a function so rewrote it to `if metadata.get("status") == "Failed":`

src/cromshell/status/command.py
bshifaw: leftover stuff from previous code. changed

src/cromshell/status/command.py
bshifaw: Good catch, changed

src/cromshell/status/command.py
bshifaw: leftover stuff from previous code. changed

src/cromshell/status/command.py
bshifaw: You're right, i removed the `workflow_failed` variable and instead the function will immediately return `True` if it encounters a failed workflow or task.

src/cromshell/status/command.py
```diff
@@ -57,23 +56,15 @@ def main(config, workflow_id):
         # TODO : Use this as a template for the Metadata subcommand
         # Get execution status count and filter the metadata down:
         request_meta_out = requests.get(
-            f"{config.cromwell_api_workflow_id}/metadata?{config.slim_metadata_parameters} "
+            f"{config.cromwell_api_workflow_id}/metadata?{config.slim_metadata_parameters}"
         )
 
         # tmp_metadata holds the workflow metadata as a dictionary
         tmp_metadata = json.loads(request_meta_out.content.decode("utf-8"))
```
SHuang-Broad, line 63: I don't think the `tmp_` prefix is necessary.

src/cromshell/status/command.py
```diff
@@ -57,23 +56,15 @@ def main(config, workflow_id):
...
-        failed = False
-        for task_status in execution_status_count:
-            for key in task_status:
-                if "Failed" in str(task_status[key]):
-                    failed = True
-                    break
 
         # Check for failures:
-        if not failed:
+        workflow_failed = check_for_failure(tmp_metadata)
```
SHuang-Broad, line 66: I'd say a variable named like `at_least_one_task_failed` better reflect what you are trying to achieve, i.e. early breaking out once a "Failed" is seen in the metadata

src/cromshell/status/command.py
```diff
@@ -114,34 +105,50 @@ def main(config, workflow_id):
...
-        # For each status for the call add the call name and status name and count
-        # to list
-        for status in set(execution_statuses):
-            status_element = f'{{"{status}": {status_count[status]}}}'
-            tmp_execution_status.append(f"{call_element}{status_element}")
-
-    tmp_execution_status_json = ", ".join(tmp_execution_status)
-    tmp_execution_status_json = "{" + tmp_execution_status_json + "}"
-    return [json.loads(tmp_execution_status_json)]
+def check_for_failure(metadata: dict):
```
SHuang-Broad, line 108: How about `check_for_task_failure`?
Then the line at 114 should be removed to caller code, i.e. this function only checks if ANY of the tasks in a workflow metadata indicates failure.

src/cromshell/status/command.py
```diff
@@ -114,34 +105,50 @@ def main(config, workflow_id):
...
+    # if status key does does not exist then we'll iterate through the dictionary
+    # value in search of a nested dictionary (indicating subworkflows)
+    # or a list which holds the task statuses.
+    for value in metadata.values():
+
+        # If a dictionary value is encountered then Depth First Search recursion
+        # is used to traverse the dictionary by iterating through the given dictionary
+        # and reporting whether it found a failure. The reason for this is due to the
+        # tree structure of the metadata dictionary which holds the task statuses
+        # in second layer dictionary key called calls.
```
SHuang-Broad, line 127: "second layer" is probably too specific.
My concern is that this code was developed with a specific example test data, so any comments that are too specific is going to confuse future developers.

SHuang-Broad (approved):
Some suggestions for small changes, mostly for naming and comments.
I believe the logic is correct now.

src/cromshell/status/command.py
bshifaw: @SHuang-Broad 

The function does check for workflow failure too. If I'm not mistaken the `status` key in a dictionary indicates the status of the workflow/subworkflow and the `executionStatus` key indicates the status of a task. Line 14 will checks for a `status` key and exits if its equal to "Failed", while Line 149 does the same for `executionStatus`. My original thinking in leaving in Line 14 was if we come across a failed `status` key for workflow which sits at a higher level in the tree compared to the `executionStatus`  then we can immediately exit without having to check through all of its tasks for a failed execution. 

Line 14 can be removed if this function should only be checking the status for tasks. Should I move forward with the name change in line removal?

src/cromshell/status/command.py
bshifaw: done

src/cromshell/status/command.py
bshifaw: replacing it to "a nested"

src/cromshell/status/command.py
```diff
@@ -114,34 +105,50 @@ def main(config, workflow_id):
...
-    tmp_execution_status_json = ", ".join(tmp_execution_status)
-    tmp_execution_status_json = "{" + tmp_execution_status_json + "}"
-    return [json.loads(tmp_execution_status_json)]
+def check_for_failure(metadata: dict):
+    """Checks a workflow metadata dictionary for failing statuses
+    Returns True to indicate workflow or some task(s) has failed"""
+
+    # If the given dictionary contains a 'status' key and has value of "Failed"
+    # then exit the function returning "True" to indicate workflow has failed
+    if metadata.get("status") == "Failed":
```
jonn-smith, line 114: I recommend moving all the status strings into an enum or similar object to reference them as objects rather than magic values.

src/cromshell/status/command.py
```diff
@@ -114,34 +105,50 @@ def main(config, workflow_id):
...
+        return True
+
+    # If the dictionary does not contain a failed value for its status key or
+    # if status key does does not exist then we'll iterate through the dictionary
```
jonn-smith, line 118: The "does not exist" part of this comment was very helpful!  I understand why you're using the `get(...)` method above now rather than the bracket operators.

src/cromshell/status/command.py
```diff
@@ -114,34 +105,50 @@ def main(config, workflow_id):
...
+                return True
+
+        # If a list value is encountered then the dictionary being traversed through is
+        # a key and value pair where the key is the task name and the value is the
+        # number of times the task was executed (shard), which is normally 1 unless the
+        # task was scattered. Hence the value for the key is a list to account
+        # for all the shards for a task. Each item(shard) within this list is a
+        # dictionary holding status for the shard.
+        # We'll want to check each shard to determine whether it's status has "Failed".
+        if isinstance(value, list):
```
jonn-smith, line 139:
```suggestion
        elif isinstance(value, list):
```

src/cromshell/status/command.py
```diff
@@ -114,34 +105,50 @@ def main(config, workflow_id):
...
+            for shard in value:
+
+                # If a key in the shared labeled "subWorkflowMetadata" then this
```
jonn-smith, line 142:
```suggestion
                # If a key in the shard is labeled "subWorkflowMetadata", then this
```

src/cromshell/status/command.py
```diff
@@ -114,34 +105,50 @@ def main(config, workflow_id):
...
+                # will another dictionary layer of subworkflow tasks that needs
```
jonn-smith, line 143:
```suggestion
                # will contain another dictionary layer of subworkflow tasks that will
```

src/cromshell/status/command.py
```diff
@@ -114,34 +105,50 @@ def main(config, workflow_id):
...
+                # to be traversed.
```
jonn-smith, line 144:
```suggestion
                # require traversal.
```

src/cromshell/status/command.py
```diff
@@ -114,34 +105,50 @@ def main(config, workflow_id):
...
+                if "subWorkflowMetadata" in shard.keys():
```
jonn-smith, line 145: Is the shard guaranteed to be a dictionary?  If not, won't this be an issue if it doesn't have a `keys()` method?

jonn-smith (changes requested):
Looks great!

I have a few suggested changes in some comments and a couple of questions.

src/cromshell/status/command.py
bshifaw: From the workflow metadata I've seen so far, yes. If we come across a key in a dictionary with a value type of list then each item in that list is always a dictionary. The items within the list represent the number of times the task was executed, it could be once or multiple items for a scattered job. Each item is dictionary which hold a key and value pair for informations like executionStatus, backendStatus, shardIndex, callRoot                          attempt.

src/cromshell/status/command.py
bshifaw: yep, since the function calls on itself repeatedly and the input that is passed doesn't not always contain a `status` key in the dictionary its better to use `get()` because it will return None instead of failing like the bracket operators when it can't find a key.

src/cromshell/status/command.py
lbergelson, line 108: I still think the function name "check_for_failure" isn't a great name.  If I check for failure is the result True if I saw a failure or true if it succeeds.  Usually functions that return a boolean should be named as a statement, not just the variable.  Unless that's not pythonic I guess.

bshifaw: Like `failure_present_in_workflow` ?

lbergelson (approved):
@bshifaw Nitpick about naming remains but I think you fixed the recursion bug. Looks good to me when you're ready.

src/cromshell/status/command.py
bshifaw: I'll change the name to `workflow_failed`

then lines 66-67 will be changed to 
```
        if not workflow_failed(metadata):
```

src/cromshell/status/command.py
bshifaw: skipping this, see line 117 thread

jonn-smith: What I meant was moving `"Failed"` into a variable that lives somewhere that you reference here instead of the hard-coded string.  It'll be useful to have the Cromwell statuses all defined somewhere.

jonn-smith (approved):

PR #150, opened 2021-02-09
Updated api_string variable

bshifaw:
Updated api string variable in the cromshell config, http utility, and status file. The update involves removing the ending "/" at initial declaration of the variable and asks functions that use the variable to add the forward slash along with their needed extension. This enables cromshell api commands like the submit function which does not use additional extensions to use the variable without having to remove the forward slash.


jonn-smith (approved):
This looks fine to me as long as there are no other places that need to have their slashes removed.

SHuang-Broad (approved):
👍

PR #148, opened 2021-02-01
Add submit function

bshifaw:
Addition of submit function allows cromshell to submit wdl and input json to cromwell server

PR #156, opened 2021-04-30
Add git actions tests

bshifaw:
Adding yml file to automate unit tests and linting through git actions

### DIFF
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -1,0 +1,36 @@
+# This is a basic workflow to help you get started with Actions
+
+name: CI
+
+# Controls when the action will run. 
+on:
+  # Triggers the workflow on push or pull request events but only for the master branch
+  push:
+    branches: [ master ]
+  pull_request:
+    branches: [ master ]
+
+  # Allows you to run this workflow manually from the Actions tab
+  workflow_dispatch:
+
+# A workflow run is made up of one or more jobs that can run sequentially or in parallel
+jobs:
+  # This workflow contains a single job called "build"
+  build:
+    # The type of runner that the job will run on
+    runs-on: ubuntu-latest
+
+    # Steps represent a sequence of tasks that will be executed as part of the job
+    steps:
+      # Checks-out your repository under $GITHUB_WORKSPACE, so your job can access it
+      - uses: actions/checkout@v2
+
+      # Runs a single command using the runners shell
+      - name: Run a one-line script
+        run: echo Hello, world!
+
+      # Runs a set of commands using the runners shell
+      - name: Run a multi-line script
+        run: |
+          echo Add other actions to build,
+          echo test, and deploy your project.
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -45,6 +45,9 @@ jobs:
       - name: Install Latest WomTool
         run: curl -s https://api.github.com/repos/broadinstitute/cromwell/releases/latest | grep "browser_download_url.*womtool" | cut -d '"' -f 4 | wget -qi -
 
+      - name: list dr
+        run: ls .
+
       - name: Create Alias for WomTool
         run: alias womtool='java -jar '${PWD}'/womtool-*'
 
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -12,7 +12,6 @@ jobs:
   build:
     # The type of runner that the job will run on
     runs-on: ubuntu-latest
-    container: bshifaw/cromshell-unit-test:latest
     strategy:
       matrix:
         python: [ 3.6, 3.7, 3.8 ]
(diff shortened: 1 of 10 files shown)


SHuang-Broad:
Q: would these CI tests pass after #148 is merged in and this branch re-based?

src/cromshell/utilities/cromshellconfig.py
```diff
@@ -151,9 +150,14 @@ def __load_cromshell_config_file(config_directory, config_file_name):
 
     cromshell_config_path = os.path.join(config_directory, config_file_name)
     if not Path(cromshell_config_path).exists():
-        LOGGER.error("Cromshell config file %s was not found", cromshell_config_path)
-        LOGGER.error("Please create %s", cromshell_config_path)
-        raise Exception(f"Cromshell config file {cromshell_config_path} was not found")
+        LOGGER.info("Cromshell config file %s was not found", cromshell_config_path)
```
bshifaw, line 153: Now creates config file instead of just failing with error

src/cromshell/utilities/cromshellconfig.py
```diff
@@ -173,6 +177,11 @@ def __get_cromwell_server(config_options: dict):
     return config_options["cromwell_server"]
 
 
+def get_cromwell_api():
```
bshifaw, line 180: Change this from string to function so that if the cromwell server gets updated after comparing workflow id with submission file, cromwell_api provides the latest cromwell server

tests/unit/test_submit.py
```diff
@@ -23,43 +23,46 @@ def temp_dir_path():
...
-        workflow_wdl_path = workflows_path.joinpath("helloWorld.wdl")
-        workflow_json_path = workflows_path.joinpath("helloWorld.json")
-
-        assert (
-            submit_command.womtool_validate_wdl_and_json(
-                wdl=str(workflow_wdl_path), wdl_json=str(workflow_json_path)
-            )
-            == 0
-        ), "Womtool should have marked valid workflow as valid."
+    womtool_path = shutil.which("womtool")
```
bshifaw, line 26: Added logic to check if womtool is present before running this test, because the git actions tests was failing since it doesn't have womtool installed

src/cromshell/submit/command.py
```diff
@@ -141,14 +141,14 @@ def validate_input(wdl: str, wdl_json: str, options_json: str, dependencies_zip:
...
 
-    womtool_path = shutil.which("womtool")
-    if womtool_path is not None:
+    if shutil.which("womtool") is not None:
 
         validation_output = None
         try:
             validation_output = subprocess.run(
-                [womtool_path, "validate", wdl, "-i", wdl_json],
-                capture_output=True,
```
bshifaw, line 151: Unit tests showed capture_output is not available in py3.6 so this was replaced with subprocess.PIPE

tests/unit/test_submit.py
bshifaw: Logic is removed and instead womtool is being installed in git action yml

.github/workflows/main.yml
```diff
@@ -0,0 +1,49 @@
+# This is a basic workflow to help you get started with Actions
```
SHuang-Broad, line 1: probably don't need this comment

.github/workflows/main.yml
```diff
@@ -0,0 +1,49 @@
...
+
+name: CI
+
+# Controls when the action will run.
+# Triggers the workflow on push or pull request events
+on: [push, workflow_dispatch]
```
SHuang-Broad, line 7: do we also want to have nightly tests?

.github/workflows/main.yml
```diff
@@ -0,0 +1,49 @@
...
+
+      - name: Install Dev Requirments
+        run: |
+          python -m pip install --upgrade pip
+          pip install -r dev-requirements.txt
+
+      - name: Install Tox
+        run: pip install tox
+
+      - name: Install Womtool
```
SHuang-Broad, line 38: minor, but do we want to move this step to before Tox intallation?

SHuang-Broad (approved):
Some minor comments.
One suggestion is to move most of the changes to the python files to a different PR, so that this PR is strictly for CI.
Otherwise you probably will have to rebase your other PR (the metadata one) as both PR modify some shared files.

.github/workflows/main.yml
bshifaw: removed

.github/workflows/main.yml
bshifaw: agreed to do this in the future. added in the cromshell sow

.github/workflows/main.yml
bshifaw: done

src/cromshell/utilities/cromshellconfig.py
```diff
@@ -151,9 +150,14 @@ def __load_cromshell_config_file(config_directory, config_file_name):
 
     cromshell_config_path = os.path.join(config_directory, config_file_name)
     if not Path(cromshell_config_path).exists():
-        LOGGER.error("Cromshell config file %s was not found", cromshell_config_path)
-        LOGGER.error("Please create %s", cromshell_config_path)
-        raise Exception(f"Cromshell config file {cromshell_config_path} was not found")
+        LOGGER.info("Cromshell config file %s was not found", cromshell_config_path)
+        LOGGER.info("Creating %s", cromshell_config_path)
+
+        config_template = '{\n"cromwell_server": "[enter server url]"\n}'
```
jonn-smith, line 156: If we have to generate a configuration file anyway, we should write out all the options with default values so the user knows what options exist by looking at the config file.

jonn-smith (changes requested):
Looks pretty good!

One request for config file generation.

src/cromshell/utilities/cromshellconfig.py
bshifaw: @jonn-smith  Done
This [variable](https://github.com/broadinstitute/cromshell/pull/156/files#diff-82107d55dd3286d108a0a9d136339c3056207f177a7d9c3d23daf3433b1c17e1R36) will hold all possible options and updated this [line](https://github.com/broadinstitute/cromshell/pull/156/files#diff-82107d55dd3286d108a0a9d136339c3056207f177a7d9c3d23daf3433b1c17e1R163) accordingly shift it from dict to json

PR #159, opened 2021-05-19
Minor change to remove temp files in the notify helper.

jonn-smith:
- This fixes an issue found where cromshell was filling /tmp disk space
with the `notify` daemon.  Not 100% sure why the trap didn't catch the
temp files, but this is a fix for the issue.


lbergelson (approved):
@jonn-smith Looks good to me.  Hopefully there's no way that makeTemp can return '*' or '/' :)

jonn-smith:
@lbergelson `makeTemp` just returns what `mktemp` would give you, so it shouldn't be an issue.  :)

PR #162, opened 2021-07-27
Removing another temporary file from notify daemon.

jonn-smith:
More cleanup for notify daemon processes.


lbergelson (approved):
@jonn-smith 👍

Issue #167, opened 2021-07-29
Timing URL is missing a colon


cwhelan:
I just installed the latest cromshell via `brew install broadinstitute/dsp/cromshell`.

When I run `cromshell timing` my browser (Chromw) tries to a URL that looks like this:

`https//~{SERVER}/api/workflows/v1/${WORKFLOW_ID}/timing`

Which is missing the colon after the `https`. The link works if I manually add a colon in.

lbergelson:
Woops.  Thanks for noticing.

PR #168, opened 2021-07-30
Fix timing command

lbergelson:
* timing was putting an extra "http" into the url which caused it to not load correctly.
* fixes https://github.com/broadinstitute/cromshell/issues/167


lbergelson:
@cwhelan This should fix the issue you found.

jonn-smith (approved):
👍

PR #169, opened 2021-08-14
Fix requests metadata call 

bshifaw:
Previously, requests call in the status command used a string constant from the cromshell utilities script to obtain the metadata for a workflow. Since the addition of the metadata command, the string constant has been updated to a dictionary. Thus when the requests line in the status command tries to use the old string it fails. In this PR the status command requests call has now been updated to properly use the constant in dictionary form.

### DIFF
--- a/src/cromshell/status/command.py
+++ b/src/cromshell/status/command.py
@@ -7,6 +7,7 @@ import click
 import requests
 
 from cromshell import log
+from cromshell.metadata import command as metadata_command
 from cromshell.utilities import cromshellconfig, http_utils, io_utils
 
 LOGGER = logging.getLogger(__name__)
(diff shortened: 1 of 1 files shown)


src/cromshell/status/command.py
```diff
@@ -57,11 +58,18 @@ def main(config, workflow_id):
     elif workflow_status == "Running":
         # Status claims this workflow is running fine, but we need to check to see
         # if there are any failed sub-processes.
-        # To do this, we use the `execution-status-count` logic with some filtering:
-        # TODO : Use this as a template for the Metadata subcommand
-        # Get execution status count and filter the metadata down:
+        # To do this, we get the workflow metadata and search for any failures
+        formatted_metadata_parameter = metadata_command.format_metadata_params(
```
bshifaw, line 62: Since this is being used in two different commands let me know if it should be relocated in the utility file, or if it is ok to leave it in the metadata command

SHuang-Broad (approved):
Looks good to me.
Were we to have integration tests, this would have been caught.
Game on! 👍

PR #170, opened 2021-08-16
Add note about bq init to README

mwalker174:
I just went through the process of getting `cost` to work. I was getting the following error at first:
```
$(standard_in) 1: parse error
(standard_in) 1: illegal character: Q
(standard_in) 1: illegal character: :
```
which was apparently due to not having a default cloud project set for `bq`. I've added a note about this in the README.


jonn-smith:
Looks good.  In addition, we should add a check to make sure this is set.

PR #172, opened 2021-08-25
Bs add ci tests

bshifaw:
- add yml to for system tests
- added submit test

### DIFF
--- a/tests/integration/test_submit.py
+++ b/tests/integration/test_submit.py
@@ -1,0 +1,25 @@
+import subprocess
+import pytest
+
+
+class TestSubmit:
+
+    def test_submit(self, local_cromwell_url: str):
+
+        result = subprocess.run(
+            [
+                'cromshell',
+                '--cromwell_url',
+                local_cromwell_url,
+                'submit',
+                'tests/workflows/helloWorld.wdl',
+                'tests/workflows/helloWorld.json'
+            ],
+            capture_output=True
+        )
+
+        assert result.returncode == 0
+
+    @pytest.fixture
+    def local_cromwell_url(self):
+        return 'http://localhost:8000'
--- a/tests/resources/cromwell_application.conf
+++ b/tests/resources/cromwell_application.conf
@@ -1,0 +1,75 @@
+include required(classpath("application.conf"))
+include "build_application.inc.conf"
+
+backend {
+  default = "Local"
+  # Do not include LocalBourneShell among the enabled backends as that will run with /bin/sh which not all tests can handle.
+  enabled = ["Local", "LocalNoDocker"]
+  providers {
+    Local {
+      actor-factory = "cromwell.backend.impl.sfs.config.ConfigBackendLifecycleActorFactory"
+      config {
+        include "local_provider_config.inc.conf"
+
+        # Have the engine (maybe) authenticate to docker.io. See BT-141 for more info.
+        include "dockerhub_provider_config_v1.inc.conf"
+      }
+    }
+    LocalNoDocker {
+      actor-factory = "cromwell.backend.impl.sfs.config.ConfigBackendLifecycleActorFactory"
+      config {
+        include "local_provider_config.inc.conf"
+        runtime-attributes = ""
+        submit-docker = null
+        kill-docker = null
+
+        # Have the engine (maybe) authenticate to docker.io. See BT-141 for more info.
+        include "dockerhub_provider_config_v1.inc.conf"
+      }
+    }
+    LocalBourneShell {
+      actor-factory = "cromwell.backend.impl.sfs.config.ConfigBackendLifecycleActorFactory"
+      config {
+        include "local_provider_config.inc.conf"
+        job-shell = /bin/sh
+
+        # Have the engine (maybe) authenticate to docker.io. See BT-141 for more info.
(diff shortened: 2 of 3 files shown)


tests/integration/test_submit.py
```diff
@@ -0,0 +1,37 @@
...
+            ("tests/workflows/helloWorld.wdl", "tests/workflows/not_valid.json", 1),
+            ("tests/workflows/not_valid.wdl", "tests/workflows/helloWorld.json", 1),
+        ]
+    )
+    def test_submit(self, local_cromwell_url: str, wdl: str, json: str, exit_code: int):
+        runner = CliRunner(mix_stderr=False)
+        result = runner.invoke(
+            cromshell,
+            [
+                "--cromwell_url",
```
lbergelson, line 21: This is a personal style thing that probably contradicts PEP-8, but when building command lines like this I like to put the arguments on the same line as the value.  I think it makes it easier to read.

ex:
```
"--cromwell_url", local_cromwell_url,
"submit",
wdl,
json
```

bshifaw: tox linting doesn't like it when they are on the same line

src/cromshell/submit/command.py
```diff
@@ -105,8 +105,9 @@ def main(config, wdl, wdl_json, options_json, dependencies_zip):
     # TODO: Refactor these file manipulations into its own "cleanup" function?
     # If we get here, we successfully submitted the job and should track it locally:
     # 1. Create a directory to hold function input files, using server name
+    server_folder_name = config.get_local_folder_name()
```
bshifaw, line 108: Instead of using a variable that was set at the initialization of cromshellconfig.py which would have been set to the default cromshell server, we now use a function to obtain the server name so that if the user-specified their server we would get their specified server instead of the default.

tests/resources/cromwell_application.conf
```diff
@@ -0,0 +1,79 @@
...
+  db {
+    driver = "com.mysql.cj.jdbc.Driver"
+    # Replaced "localhost" with "172.17.0.2" to allow cromwell container to
+    # communicate with mysql container on docker network
+    url = "jdbc:mysql://172.17.0.2/DatabaseName?rewriteBatchedStatements=true&useSSL=false"
+    user = "ChooseAName"
+    password = "YourOtherPassword"
+    connectionTimeout = 5000
+  }
+}
```
bshifaw, line 79:
```suggestion
}

```

.github/workflows/integration_tests.yml
```diff
@@ -0,0 +1,74 @@
...
+
+          # Give cromwell server time to setup
+          sleep 1m
+
+      - name: Install Tox
+        run: pip install tox
+
+      - name: Run Tox integration tests
+        env:
+          TMPDIR: /tmp/ # Used to create cromshell config dir in cromshellconfig.py
```
bshifaw, line 72: make cromshell configurable via environment instead.

tests/integration/test_submit.py
```diff
@@ -0,0 +1,111 @@
...
+from traceback import print_exception
+
+import pytest
+from click.testing import CliRunner
+
+from cromshell.__main__ import main_entry as cromshell
+
+
+class TestSubmit:
+    workflow_id = None  # Anyway to avoid using global variables?
```
bshifaw, line 14:
```suggestion
    workflow_id = None 
```

lbergelson, line 14: Remove this if it's no longer needed.

tests/integration/test_submit.py
```diff
@@ -0,0 +1,111 @@
...
+            assert result.exit_code == exit_code, (
+                f"\nSTDOUT:\n{result.stdout}"
+                f"\nSTDERR:\n{result.stderr}"
+                f"\nExceptions:\n{result.exception}"
+                f"\n{print_exception(*result.exc_info)}"
+            )
+
+            # If submission passed get workflow id and set workflow id global variable
+            if exit_code == 0:
+                # Using regex to match json output from stdout
```
bshifaw, line 52: Use hide turtle parameter

tests/integration/test_submit.py
```diff
@@ -0,0 +1,111 @@
...
+                TestSubmit.workflow_id = stdout_substring_formatted["id"]
+
+    def test_workflow_id_in_db(
+        self, local_hidden_cromshell_folder, local_workflow_database_tsv
+    ):
+        test_workflow_id = TestSubmit.workflow_id
+        local_db_file = local_workflow_database_tsv
+
+        # opening a text file
+        file1 = open(local_db_file, "r")
```
bshifaw, line 73: open file `with`

tests/integration/test_submit.py
```diff
@@ -0,0 +1,111 @@
...
+                    ).group(0)
+                except AttributeError:
+                    print("Trouble extracting json in stdout")
+                    print(repr(result.stdout))
+                    print(result_stdout_substring)
+
+                stdout_substring_formatted = json.loads(result_stdout_substring)
+                TestSubmit.workflow_id = stdout_substring_formatted["id"]
+
+    def test_workflow_id_in_db(
```
bshifaw, line 66: Good to combine tests here, to avoid global variables with tests that are linked.

PR #183, opened 2021-11-05
removed dataclasses in setup.py 

bshifaw:
Removed dataclasses from the setup.py because starting with python3.7, the package is included with python releases. ([reference](https://pypi.org/project/dataclasses/)). Cromshell only supports >=3.7


SHuang-Broad (approved):
👍

Issue #139, opened 2020-10-01
WDL validation fails when it requires a dependency zip


tnguyensanger:
When womtool is in the PATH, the womtool is invoked, forwarding the WDL and input json:

https://github.com/broadinstitute/cromshell/blob/c7ea8837c0739c911ceeaa00746461c535adb97a/cromshell#L772

If a dependencies zip has been passed to cromshell, it is not forwarded to womtool, causing WDL validation to fail.

Seen in cromshell v0.4.3

jonn-smith:
@tnguyensanger Thanks for the report.  This is a good finding and kind of an oversight on my/our part.

I'll see if we can bang out a quick fix.

kshakir:
As of Womtool 78 can one can specify a dependencies zip during validation?

The wiring into the Womtool REST API seems straightforward and FYI would need Jira tickets, compliance review, tests, etc. If someone wants to try adding dependencies to the REST API a shot I'm happy to provide guidance.

Seemingly not accepting dependencies for validation:
- [Womtool CLI](https://github.com/broadinstitute/cromwell/blob/78/womtool/src/main/scala/womtool/cmdline/PartialWomtoolCommandLineArguments.scala#L5-L11)
- [Womtool REST API](https://github.com/broadinstitute/cromwell/blob/78/engine/src/main/scala/cromwell/webservice/routes/WomtoolRouteSupport.scala#L58)

PR #196, opened 2022-04-28
Use the server to run womtool validation

kshakir:


kshakir:
- Added an option to disable validation due to the validation false negatives reported in #139 
- Moved tests that were calling out to the `womtool` executable from unit to integration tests
- Integration tests for validation now make calls to the cromwell server

src/cromshell/utilities/cromshellconfig.py
```diff
@@ -23,7 +23,8 @@
     "subWorkflowMetadata",
     "subWorkflowId",
 ]
-API_STRING = "/api/workflows/v1"
+CROMWELL_API_STRING = "/api/workflows/v1"
+WOMTOOL_API_STRING = "/api/womtool/v1"
```
kshakir, line 27: One of many examples where I am open to naming suggestions. For example: here the names follow the CLI naming scheme (`cromwell` and `womtool`), but I could also see naming them after the endpoints (`workflows` and `womtool`).

src/cromshell/submit/command.py
```diff
@@ -136,38 +151,54 @@ def validate_input(wdl: str, wdl_json: str, options_json: str, dependencies_zip:
         io_utils.assert_file_is_not_empty(dependencies_zip, "Dependencies Zip")
 
     # At this point, we should validate our inputs if womtool is in PATH:
```
lbergelson, line 153: this comment is out of date now

src/cromshell/utilities/cromshellconfig.py
lbergelson: I think cromwell and womtool are very understandable.  I'm definitely used to it being called cromwell, not "workflows api"

lbergelson (commented):
@kshakir This seems sane to me.  It nice to us the server womtool so it stays in sync. I found 1 tiny issue, and I have few questions.

1.  The reason we want to use validate before submitting is to reduce latency and get an instant result.  Does this introduce noticeable latency to the process?  

2. Is there a way to validate it WITH the dependencies? 

3.  If there is no way to validate with the dependencies zip, and it will ensure that it fails, can we make it automatically skip and notify if there is one included.

src/cromshell/submit/command.py
```diff
@@ -136,38 +151,54 @@ def validate_input(wdl: str, wdl_json: str, options_json: str, dependencies_zip:
...
-    womtool_validate_wdl_and_json(wdl, wdl_json)
+    womtool_validate_wdl_and_json(wdl, wdl_json, config)
 
 
-def womtool_validate_wdl_and_json(wdl: str, wdl_json: str):
+def womtool_validate_wdl_and_json(
```
lbergelson, line 157: Lets make this -> None instead of always returning 0.

kshakir:
> * The reason we want to use validate before submitting is to reduce latency and get an instant result.  Does this introduce noticeable latency to the process?

I don't notice significant latency. My guess is that the time to communicate over HTTPS should be about the same latency as spinning up a JVM. So I would expect 0-to-1 second latencies for both the womtool CLI or REST API.

> * Is there a way to validate it WITH the dependencies?

Not yet as far as I can tell. I added a comment with a pointer in #139 where one could wire in the dependencies zip in the Womtool (aka Cromwell) codebase with patience.

> * If there is no way to validate with the dependencies zip, and it will ensure that it fails, can we make it automatically skip and notify if there is one included.

Added. As always feel free to refactor, edit inglish, etc.

lbergelson (approved):
@kshakir Looks good to me.

PR #203, opened 2022-04-29
Add `update-server` command to 2.0

### LINKED ISSUE #199: `update-server` command
Port the `update-server` command over from 1.0

Allows the user to programmatically change the server name in the config file, without having to know how to do this manually.

sjfleming:
Command that will update the default Cromwell server for future workflow submissions.

Closes #199

### DIFF
--- a/src/cromshell/update_server/command.py
+++ b/src/cromshell/update_server/command.py
@@ -2,6 +2,7 @@ import json
 import logging
 import os
 from pathlib import Path
+from typing import Dict
 
 import click
 
@@ -15,6 +16,8 @@ CROMWELL_SERVER_KEY = "cromwell_server"
 def main(config, cromwell_server_url):
     """Update the default Cromwell server in the config file"""
 
+    LOGGER.info("update-server")
+
     # location of the config file
     cromshell_config_path = os.path.join(
         config.config_dir, config.CROMSHELL_CONFIG_FILE_NAME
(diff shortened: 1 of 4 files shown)


src/cromshell/update_server/command.py
```diff
@@ -0,0 +1,41 @@
...
+    # location of the config file
+    cromshell_config_path = os.path.join(
+        config.config_dir, config.CROMSHELL_CONFIG_FILE_NAME
+    )
+    assert os.access(
+        cromshell_config_path, mode=os.W_OK
+    ), f"Cannot write to Cromshell config file {cromshell_config_path}"
+
+    # the contents of the config file as a dict
+    cromshell_config_dict = config.cromshell_config_options
```
bshifaw, line 27: It isn't [good practice](https://docs.quantifiedcode.com/python-anti-patterns/readability/putting_type_information_in_a_variable_name.html) to use type suffix when naming variables, it would be best to rename `cromshell_config_dict`

src/cromshell/update_server/command.py
```diff
@@ -0,0 +1,41 @@
...
+
+    # update the cromwell server
+    if CROMWELL_SERVER_KEY not in cromshell_config_dict.keys():
+        cromshell_config_dict.update({CROMWELL_SERVER_KEY: cromwell_server_url})
+    else:
+        cromshell_config_dict[CROMWELL_SERVER_KEY] = cromwell_server_url
+
+        # write the modified config file
```
bshifaw, line 35: Would this part need to be tab-shifted to the left so that it updates the cromshell config file all the time, and not just when it finds a cromwell_server_key in the cromshell_config_options?

src/cromshell/update_server/command.py
```diff
@@ -0,0 +1,41 @@
...
+LOGGER = logging.getLogger(__name__)
+CROMWELL_SERVER_KEY = "cromwell_server"
+
+
+@click.command(name="update-server")
+@click.argument("cromwell_server_url", required=True, nargs=1)
+@click.pass_obj
+def main(config, cromwell_server_url):
+    """Update the default Cromwell server in the config file"""
+
```
bshifaw, line 17:
```suggestion
LOGGER.info("update-server")

```

bshifaw (changes requested):
Some minor changes

src/cromshell/update_server/command.py
sjfleming: Haha, okay I tend to to that, but looks like you're right.  What name would you recommend?  Happy to use whatever.  I guess I could do this 
```python
cromshell_config: Dict = config.cromshell_config_options
```
though I think it looks kind of strange

src/cromshell/update_server/command.py
sjfleming: You are totally right, I just saw that last night and converted this back to a draft PR so I wouldn't forget to fix that, good catch.

src/cromshell/update_server/command.py
sjfleming: Will do

src/cromshell/update_server/command.py
sjfleming: I'm going to go with that for now.  I want somebody who reads the code to understand that `config.cromshell_config_options`, which could be anything, is actually a Dict

src/cromshell/update_server/command.py
bshifaw: It "shouldn't" be missing a `cromwell_server`. On the initial run, cromshell will look for the config json and if it doesn't find it, it will create one from a template that has the `cromwell_server` key. Then it will ask the user to update its value. It's possible the user could create their own or edit a config file themselves and remove the key, this would cause cromshell to fail since it won't be able to access the Cromwell server. In the long run, I'd like to add some validation scripts for the config file to make sure certain keys are available.

bshifaw (approved):
Looks good

sjfleming:
Thanks @bshifaw .  Let me know what to do about merging.  Will you do that?  Seems like if we all try to merge at once without rebasing on each others' merges, there could be problems.

src/cromshell/update_server/command.py
bshifaw: Nice, I didn't know you could specify the data structure of a variable in that way

sjfleming: Well... it's a "type hint"... so it's not enforced in any way.  It's just a hint for the reader (and the python IDE), but it can be helpful in a case like this (even though it looks a bit weird)

bshifaw:
Feel free to merge after rebasing

PR #205, opened 2022-04-29
Refactor status

SHuang-Broad:
So that other commands can reuse its logic (notably, `notify`).

### DIFF
--- a/src/cromshell/status/command.py
+++ b/src/cromshell/status/command.py
@@ -52,9 +52,7 @@ def main(config, workflow_id):
     return ret_val
 
 
-def query_status(
-    config, workflow_id: str, log_to_screen: bool = True
-) -> (str, str, int):
+def query_status(config, workflow_id: str) -> (str, str, int):
     """
 
     :param config: object holding configurations
@@ -82,9 +80,8 @@ def query_status(
         + cromshellconfig.WorkflowStatuses.Aborted.value
     ):
         ret_val = 1
-        if log_to_screen:
-            log.display_logo(io_utils.dead_turtle)
-    elif workflow_status == "Running":
+        log.display_logo(io_utils.dead_turtle)
+    elif workflow_status == cromshellconfig.WorkflowStatuses.Running.value[0]:
         # Status claims this workflow is running fine, but we need to check to see
         # if there are any failed sub-processes.
         # To do this, we get the workflow metadata and search for any failures
@@ -108,11 +105,9 @@ def query_status(
         if not workflow_failed(metadata):
             # We could not find 'Fail' in our metadata, so our
             # original Running status is correct.
-            if log_to_screen:
-                log.display_logo(io_utils.turtle)
+            log.display_logo(io_utils.turtle)
         else:
-            if log_to_screen:
-                log.display_logo(io_utils.doomed_logo)
+            log.display_logo(io_utils.doomed_logo)
             workflow_status = cromshellconfig.WorkflowStatuses.DOOMED.value[0]
             message = (
                 "The workflow is Running but one of the instances "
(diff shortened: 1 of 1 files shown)


src/cromshell/status/command.py
```diff
@@ -32,19 +32,47 @@ def main(config, workflow_id):
     # Check if Cromwell Server Backend works
     http_utils.assert_can_communicate_with_server(config)
 
+    requested_status_json, workflow_status, updated_ret_val = query_status(
+        config, workflow_id
+    )
+    ret_val = updated_ret_val
+
+    # Display status to user:
+    line_string = requested_status_json
```
lbergelson, line 41: why rename this here?  Lets just print the requested_status_json after replacement.

src/cromshell/status/command.py
```diff
@@ -53,7 +81,7 @@ def main(config, workflow_id):
     ):
         ret_val = 1
         log.display_logo(io_utils.dead_turtle)
-    elif workflow_status == "Running":
+    elif workflow_status == cromshellconfig.WorkflowStatuses.Running.value[0]:
```
lbergelson, line 84: This is weird.  I feel like there must be a better solution but we don't have to find it right now.

lbergelson (approved):
@SHuang-Broad Looks fine.  Minor comments.  At some point we might want to move it to utils if people are going to call into it a lot.

src/cromshell/status/command.py
```diff
@@ -32,19 +32,47 @@ def main(config, workflow_id):
...
     http_utils.assert_can_communicate_with_server(config)
 
+    requested_status_json, workflow_status, updated_ret_val = query_status(
+        config, workflow_id
+    )
+    ret_val = updated_ret_val
+
+    # Display status to user:
+    line_string = requested_status_json
+    print(line_string.replace(",", ",\n"))
```
bshifaw, line 42: Its fine as is but would be nice to use [pretty_print_json()](https://github.com/broadinstitute/cromshell/blob/ae1aa5c402335517779c529641c3b0c35b69e0e7/src/cromshell/utilities/io_utils.py#L113)

src/cromshell/status/command.py
```diff
@@ -32,19 +32,47 @@ def main(config, workflow_id):
...
+        workflow_database_path=config.submission_file_path,
+        workflow_id=workflow_id,
+        column_to_update="STATUS",
+        update_value=workflow_status,
+    )
+
+    return ret_val
+
+
+def query_status(config, workflow_id: str) -> (str, str, int):
```
bshifaw, line 55: Not now but later, would be good to add unit tests

bshifaw (approved):

src/cromshell/status/command.py
bshifaw: +1
Plus It isn't [good practice](https://docs.quantifiedcode.com/python-anti-patterns/readability/putting_type_information_in_a_variable_name.html) to use type suffix when naming variables

SHuang-Broad: Push back a little: I didn't change this as it was in the original code. I used pycharm to refactor out a function. 
Regardless, I'll update this PR with the requested changes, and move this to utils.

SHuang-Broad (commented):
Some comments, let's make this PR correct and not merge yet.

src/cromshell/status/command.py
lbergelson: It feels like there should be a less awkward way of matching against the enum values.  We shouldn't have to decompose the list manually. Maybe we should overload the equality option in the enum to check all the list values so we don't have to do this awkward thing.

PR #206, opened 2022-04-29
added an optional utility to download and wrap womtool

gileshall:
This is groundwork to make it easy to execute cromshell from a docker image

### DIFF
--- a/utils/womtool.py
+++ b/utils/womtool.py
@@ -1,0 +1,45 @@
+#!/usr/bin/env python
+# handy utility to download and wrap womtool
+
+import os
+import sys
+import subprocess
+from requests import get
+from shutil import which
+from pathlib import Path
+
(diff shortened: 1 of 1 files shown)


lbergelson:
This seem like it could be useful for people who aren't using brew or conda.  If we merge #196 it would preempt the need for this by pushing the womtools validation to the server.  @kshakir @gileshall

lbergelson:
This seem like it could be useful for people who aren't using brew or conda.  If we merge #196 it would preempt the need for this by pushing the womtools validation to the server.  @kshakir @gileshall

lbergelson:
@gileshall I merged #196 so it should no longer be relevant to install womtool locally.  I'm going to close this since i think it's no longer necessary.  It was a good idea though!

PR #207, opened 2022-04-29
Zip the --dependencies-zip arg when passed as a dir (Closes #193)

### LINKED ISSUE #193: submit command could go one step further
Making the dependency zip for the user.

Who hasn't been bitten by their own mistakes from forgetting to remake the zip after changing the imported WDLs?

kshakir:


src/cromshell/utilities/io_utils.py
```diff
@@ -145,6 +149,39 @@ def create_directory(
         raise
 
 
+def temporary_zip_directory(zip_or_directory: Optional[str]) -> Optional[str]:
+    """Return the path to a temporary zip file containing contents of the directory,
+    otherwise return the unmodified zip_or_directory argument."""
+
+    if zip_or_directory and Path(zip_or_directory).is_dir():
+        zip_base = temporary_path(Path(zip_or_directory).name)
```
bshifaw, line 157: To avoid having to create and clean files from the filesystem, would it make sense to use pythons [zipfile](https://docs.python.org/3/library/zipfile.html#zipfile-objects) to create a zipped file-like object? It is used in the unit tests but not here, how come?

kshakir: Created #214 with one option for in-memory submission.

The implementation in this PR made sure that bit-for-bit what was sent to Cromwell was later copied to the `run_directory`.

I also used the higher-level `shutil` to write the on-disk ZipFile in the main code because I wasn't confident in writing the files correctly. But the only way to read a zip in the test code (afaict) was to use the `ZipFile`.

kshakir:
Closed in favor of the above mentioned in-mem version.

PR #213, opened 2022-05-04
Update workflow database to tab-delimited format if necessary

### LINKED ISSUE #212: Old database files incompatible with 2.0
`
Using a cromshell1.0 all.database.tsv, seems like 1.0 uses spaces while 2.0 uses tabs to separate the header, this caused cromshell2 to error out.
`

There's an issue where if you used an old enough version of Cromshell 1.0, then the beginning of your submission database will be delimited by spaces rather than tabs.  This will cause Cromshell 2.0 to error on any command that needs to read the database file.

sjfleming:
Fixes #212

### DIFF
--- a/src/cromshell/utilities/cromshellconfig.py
+++ b/src/cromshell/utilities/cromshellconfig.py
@@ -101,6 +101,10 @@ def resolve_cromwell_config_server_address(server_user=None, workflow_id=None):
             "Checking submission file for associated cromwell server with the provided "
             "workflow id."
         )
+
+        # perform a check to see if the submission database is in a old format without tabs
+        __ensure_correct_submission_database_format(submission_file_path)
+
         with open(submission_file_path, "r") as csv_file:
             reader = csv.DictReader(csv_file, delimiter="\t")
             id_in_file = False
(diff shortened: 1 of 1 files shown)


jamesemery:
@sjfleming I just tested and this worked great and appears to have fixed my borken list file

bshifaw:
@[sjfleming](https://github.com/sjfleming) thanks so much for working on this. Mind moving the function to [submissions_file_utils.py](https://github.com/broadinstitute/cromshell/blob/sf_fix_workflow_database_2.0/src/cromshell/utilities/submissions_file_utils.py) so that all submission file database related functions is in one place. Then move the function call to the main file right before the [resolve_cromwell_server](https://github.com/broadinstitute/cromshell/blob/64c12c89ed1ad187baee78046bb9f5943af0b9ee/src/cromshell/__main__.py#L97) call, that way the function doesn't get overlooked in a function focused on the Cromwell server.

src/cromshell/utilities/cromshellconfig.py
```diff
@@ -122,6 +127,32 @@ def resolve_cromwell_config_server_address(server_user=None, workflow_id=None):
...
+    then update the database so it is tab-delimited."""
+
+    old_format = False
+
+    with open(submission_file_path, "r") as f:
+        first_line = f.readline()
+        if "\t" not in first_line:
+            old_format = True
+
+    if old_format:
```
lbergelson, line 143: This has the problem that if I start 2 cromshells at the same time it can easily clobber itself.  Although that's sort of a general problem with how cromshell works really....

It's often best to do these operations so the update is atomicish in case it fails midway through it doesn't destroy the db.  Usually the pattern is read -> copy into new file -> move new file into old file.  I think @bshifaw knows a magic library that makes that automatic with an "update" file option that handles it all in the background.

lbergelson (commented):
@sjfleming one comment on this.  Looks like a good idea though!

src/cromshell/utilities/cromshellconfig.py
sjfleming: This is a really good point

sjfleming:
I'll try to rebase and then make those changes @bshifaw , and I'll also use that non-clobbering pattern @lbergelson .  I'll just implement the pattern myself for now unless Beri knows what that library is.

sjfleming:
@bshifaw FYI, it looks like `isort` disagrees with `black` a lot of the time, and so sometimes linting becomes a real pain.  Do you need `isort`?

sjfleming:
I updated to @lbergelson 's method of file over-writing, and I create a backup file for good measure.

Should be ready to merge!  I will leave that to you @bshifaw

bshifaw (approved):
Looks good!

PR #214, opened 2022-05-06
Zip the --dependencies-zip arg in memory when passed as a dir

kshakir:


kshakir:
@bshifaw Some differences between the current two options:

In #207 
- the dependency-directory is stored in a temporary zip file
- the zip file is loaded using the existing code for processing dependency_zips
- the zip that was submitted to Cromwell is copied to the `run_directory`
- the temporary directory is cleaned up

Here in this PR, currently:
- the dependency-directory is put into an in-memory zip and returned as a `BytesIO`
- the `BytesIO` is passed to Cromwell the same as the WDL and input-json `BinaryIO` objects
- the original dependency-directory is copied to the `run_directory`
- there is no temporary directory to clean up

Open to other ideas. For example one could have cromshell still zip in-memory for submission to Cromwell, and afterwards write the zipped output to the `run_directory` instead of copying the unpacked depedpendency-directory.

PR #215, opened 2022-05-09
Fix brew uninstallation instructions

cjllanwarne:


bshifaw (approved):
Thanks for catching this

Issue #217, opened 2022-07-29
Missing requirement: `gcsfs`


sjfleming:
This line imports `gcsfs`, but it is not in `requirements.txt`:

https://github.com/broadinstitute/cromshell/blob/634803e3b07a41ca603629e87162952588dfcb32/src/cromshell/logs/command.py#L5

This makes a `pip install` fail when I try

```console
git clone https://github.com/broadinstitute/cromshell.git
cd cromshell
git checkout cromshell_2.0
pip install .
```

PR #218, opened 2022-07-29
Add `gcsfs` to requirements.txt

sjfleming:
@jamesemery , should there be a specific version required?

### DIFF
--- a/requirements.txt
+++ b/requirements.txt
@@ -2,3 +2,4 @@ termcolor>=1.1.0
 click>=8.0.0
 requests>=2.27.1
 pygments>=2.12.0
+gcsfs


sjfleming:
Closes #217

jamesemery:
@sjfleming heh looks like this IS my fault... I am not an expert in python package management so I can't say if it should matter what version we peg it to but if we do we should make sure to update it in both this place and in dev-requirements.txt.

SHuang-Broad:
Should there be a test for `requirements.txt`?
E.g. `pip check` in the test infrastructure?

sjfleming:
@jamesemery oh, I see you already have it included in `dev-requirements.txt`.

To be honest, I'm not sure I see where `dev-requirements.txt` is ever used?  I must be missing the reference to it from somewhere

sjfleming:
@SHuang-Broad I didn't know about `pip check`.   Sounds interesting

sjfleming:
I'll wait for @bshifaw 's input before I merge

bshifaw:
Yes! thanks, good to merge

bshifaw (approved):

PR #221, opened 2022-10-07
scripts to publish to pypi

bshifaw:
Adding scripts to publish cromshell to pypi
Changed call from "cromshell-alpha" to "cromshell-beta"
Updated Readme to show what commands are available in cromshell 2

### DIFF
--- a/.github/workflows/pypi_packaging.yml
+++ b/.github/workflows/pypi_packaging.yml
@@ -1,0 +1,33 @@
+name: Publish to PyPi
+
+# Controls when the action will run.
+# Triggers the workflow on push or pull request events
+on: [push]
+
+# A workflow run is made up of one or more jobs that can run sequentially or in parallel
+jobs:
+  # This workflow contains a single job called "build"
+  build:
+    # The type of runner that the job will run on
+    runs-on: ubuntu-latest
+
+    # Steps represent a sequence of tasks that will be executed as part of the job
+    steps:
+      # Checks-out your repository under $GITHUB_WORKSPACE, so your job can access it
+      - uses: actions/checkout@v2
+
+      - name: Setup Python
+        uses: actions/setup-python@v2
+        with:
+          python-version: '3.10'
+
+      - name: Setup Env
+        run: |
+          python3 -m pip install --upgrade pip
+          python3 -m pip install --upgrade twine
+
+      - name: Build
+        run: python3 -m build
+
+      - name: Publish to PyPi
+        run: python3 -m twine upload --username __token__ --password ${{ secrets.PYPI_TOKEN }} --repository testpypi dist/*
--- a/.github/workflows/pypi_packaging.yml
+++ b/.github/workflows/pypi_packaging.yml
@@ -23,8 +23,7 @@ jobs:
 
       - name: Setup Env
         run: |
-          python3 -m pip install --upgrade pip
-          python3 -m pip install --upgrade twine
+          python3 -m pip install --upgrade pip twine build
 
       - name: Build
         run: python3 -m build
--- a/.github/workflows/pypi_packaging.yml
+++ b/.github/workflows/pypi_packaging.yml
@@ -1,8 +1,21 @@
 name: Publish to PyPi
 
 # Controls when the action will run.
-# Triggers the workflow on push or pull request events
-on: [push]
+# Triggered manually
+on:
+  workflow_dispatch:
+    PyPI_Repo:
+        description: 'Where to upload Package'
+        required: true
+        default: 'testpypi'
+        type: choice
+        options:
+        - testpypi
+        - pypi
+    version:
+      description: 'Release Version'
+      required: true
+      type: string
 
 # A workflow run is made up of one or more jobs that can run sequentially or in parallel
 jobs:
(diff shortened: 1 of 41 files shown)


setup.py
```diff
@@ -21,14 +26,14 @@
...
-    tests_require=["coverage", "pytest"],
+    tests_require=test_install_requires,
     python_requires=">=3.7",
     packages=find_packages("src"),
     package_dir={"": "src"},
     classifiers=[
-        "Development Status :: 3 - Alpha",
+        "Development Status :: 3 - Beta",
         "Intended Audience :: Science/Research",
-        "License :: OSI Approved :: BSD 3-Clause",
```
bshifaw, line 31: there is no 3-Clause classifier in pypi

.github/workflows/pypi_packaging.yml
```diff
@@ -0,0 +1,32 @@
+name: Publish to PyPi
+
+# Controls when the action will run.
+# Triggers the workflow on push or pull request events
+on: [push]
```
bshifaw, line 5: This needs to be changed so that it only runs when pushed to cromsell2.0. 
keeping it as is for now so that I can test this action once the pypi token is added to git secrets

.github/workflows/pypi_packaging.yml
```diff
@@ -0,0 +1,32 @@
...
+
+      - name: Setup Env
+        run: |
+          python3 -m pip install --upgrade pip twine build
+
+      - name: Build
+        run: python3 -m build
+
+      - name: Publish to PyPi
+        run: python3 -m twine upload --username __token__ --password ${{ secrets.PYPI_TOKEN }} --repository testpypi dist/*
```
bshifaw, line 32: `testpypi` needs be changed to `pypi` before merging to `cromshell2` branch. Leaving as is for now for testing

bshifaw, line 32: Explain what tokens are

.github/workflows/pypi_packaging.yml
bshifaw: Keep this manual. 
use testpypi as a dev release

bshifaw: Will change to [workflow_dispatch](https://docs.github.com/en/actions/managing-workflow-runs/manually-running-a-workflow) once ready for merging.

README.md
```diff
@@ -102,7 +110,7 @@ Current version: 2.0.0
 From brew
 
     brew tap broadinstitute/dsp
-    brew install cromshell@2.0.0.alpha.1
+    brew install cromshell@2.0.0.beta
```
lbergelson, line 113: lets make this beta.1 in case we need a beta 2

pyproject.toml
```diff
@@ -0,0 +1,14 @@
+[build-system]
+requires = ["setuptools>=61.0"]
+build-backend = "setuptools.build_meta"
+
+[project]
+# TODO: update name (cromshell) and version (2.0.0.beta.1)
```
lbergelson, line 6: what do we do with this todo?  is this handled by the sed lines in github actions?

setup.py
```diff
@@ -5,6 +5,11 @@
 with open("requirements.txt") as fh:
     install_requires = fh.readlines()
 
+with open("test-requirements.txt") as fh:
+    test_install_requires = fh.readlines()
+    test_install_requires.remove("-r requirements.txt")
```
lbergelson, line 10: What's this doing?

setup.py
```diff
@@ -21,14 +26,14 @@
...
     long_description=long_description,
     install_requires=install_requires,
-    tests_require=["coverage", "pytest"],
+    tests_require=test_install_requires,
     python_requires=">=3.7",
     packages=find_packages("src"),
     package_dir={"": "src"},
     classifiers=[
-        "Development Status :: 3 - Alpha",
+        "Development Status :: 3 - Beta",
```
lbergelson, line 34: this should be 4 - Beta shouldn't it?

lbergelson (changes requested):
@bshifaw  A few minor comments then 👍

.github/workflows/pypi_packaging.yml
```diff
@@ -0,0 +1,59 @@
...
+
+      - name: Setup Env
+        run: |
+          python3 -m pip install --upgrade pip twine build
+
+      - name: Build
+        run: python3 -m build
+
+      - name: Replace version tag in pyproject.toml
+        run: sed -iE "s/version = $PARTITION_COLUMN.*/version = \"${{ inputs.version }}\"/" pyproject.toml
```
bshifaw, line 45: `$PARTITION_COLUMN` might not needed

README.md
```diff
@@ -10,24 +10,32 @@
 # cromshell
  A script for submitting workflows to a cromwell server and monitoring / querying their results.
 
-requires `column`, `curl`, `mail`, and [jq](https://stedolan.github.io/jq/)
-
-Current version: 2.0.0 
+Current version: 2.0.0.beta
```
bshifaw, line 13: Add somewhere why cromshell 2 is amazing.

setup.py
bshifaw: add comment

README.md
```diff
@@ -10,24 +10,32 @@
...
-         cromshell-alpha -t 20 metadata
-         cromshell-alpha logs -2
+         cromshell-beta submit workflow.wdl inputs.json options.json dependencies.zip
+         cromshell-beta status
+         cromshell-beta -t 20 metadata
+         cromshell-beta logs -2
 ```
 
 ## Supported Flags:
+  * `--hide_logo`
```
lbergelson, line 25:
```suggestion
  * `--no_turtle`. or `--🐢🚫` --I_hate_turtles
```

lbergelson: We should support all of those options.  Unfortunately it doesn't seem like any fonts support the combining cancellation character with the turtle emoji which is really the ideal name.

.github/workflows/pypi_packaging.yml
bshifaw: `sed -iE "s/^\s*version\s*=\s*.*$/version = \"${{ inputs.version }}\"/" pyproject.toml`

setup.py
bshifaw: Its putting the packages in the "test-requirements.txt" into a list, but then removes "-r requirements.txt". However, the packages in requirements.txt is added in the next line `test_install_requires.extend(install_requires)`. All this is because pypi doesn't know how to read a given file to retrieve packages.

.github/workflows/pypi_packaging.yml
bshifaw: leaving testpypi in for draft releases

PR #229, opened 2022-11-28
Add unit test for test_io_utils.py (merge to cromshell_2.0)

eviewan:
I added a unit test for testing whether a directory is empty.

### DIFF
--- a/tests/unit/test_io_utils.py
+++ b/tests/unit/test_io_utils.py
@@ -44,6 +44,13 @@ class TestIOUtilities:
                 path=empty_temp_file_path, description="Io Utils"
             ), "Provided a fake file path, function is fail"
 
+        if len(os.listdir(tmp_path)) == 0:
+            with pytest.raises(EOFError):
+                io_utils.assert_path_is_not_empty(
+                    path=tmp_path, description="Io Utils"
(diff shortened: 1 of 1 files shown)


tests/unit/test_io_utils.py
```diff
@@ -44,6 +44,13 @@ def test_assert_path_is_not_empty(self, tmp_path):
                 path=empty_temp_file_path, description="Io Utils"
             ), "Provided a fake file path, function is fail"
 
+        if len(os.listdir(tmp_path)) == 0:
```
bshifaw, line 47: Since this line is testing whether the function raises an error if the provided directory path is empty it might be better placed before line 38. Because line 38 the `empty.txt` file is created in the tmp folder.

tests/unit/test_io_utils.py
```diff
@@ -44,6 +44,13 @@ def test_assert_path_is_not_empty(self, tmp_path):
                 path=empty_temp_file_path, description="Io Utils"
             ), "Provided a fake file path, function is fail"
```
bshifaw, line 45: Mind fixing the comment to `"Provided path to an empty file, function expected to fail"`
and the same with the ones above

PR #232, opened 2022-12-05
added update-server description to README.md

bshifaw:
added update-server description to README.md


eviewan (approved):
this looks good 
(question - in what situation would one wants to update cromwell server? what's the benefit of doing that?)

bshifaw:
Thanks for the review. 
If a user is working with different cromwell servers or want to use new cromwell server from the latest release then they would need to update which one cromshell is sending jobs to.

PR #244, opened 2023-02-10
Linting fixes

bshifaw:


lbergelson (approved):
If this is what the linter wants, this is what it shall have!

PR #248, opened 2023-03-16
Add list-outputs

bshifaw:
Add list-outputs command
Add unit and integrations tests


tests/integration/utility_test_functions.py
```diff
@@ -9,10 +9,11 @@
 from cromshell.utilities import cromshellconfig
 
 
-def run_cromshell_command(command: list, exit_code: int):
+def run_cromshell_command(command: list, exit_code: int, options: list = None):
```
SHuang-Broad, line 12: Let's document what these parameters are.

tests/integration/utility_test_functions.py
```diff
@@ -24,12 +25,18 @@ def run_cromshell_command(command: list, exit_code: int):
     :return: results from execution
     """
 
+    command_with_options = command.copy()
+    if options:
+        for item in options:
+            command_with_options.insert(1, item)
```
SHuang-Broad, line 31: what if the options are meant to be in order? This seems to reverse that.

tests/unit/test_list_outputs.py
```diff
@@ -0,0 +1,175 @@
...
+
+import pytest
+
+from cromshell.list_outputs import command as list_outputs_command
+
+
+class TestListOutputs:
+    """Test the execution list-outputs command functions"""
+
+    # def get_workflow_level_outputs(config) -> dict:
```
SHuang-Broad, line 11: is this a self-reminder? If so, let's be a bit explicit in the comments.

src/cromshell/list_outputs/command.py
```diff
@@ -0,0 +1,200 @@
...
+        indent (bool): Whether to indent the output"""
+
+    i = "\t" if indent else ""
+
+    if isinstance(output_value, str):
+        if is_path_or_url_like(output_value):
+            print(f"{i}{output_name}: {output_value}")
+
+
+def is_path_or_url_like(in_string: str) -> bool:
```
SHuang-Broad, line 186: I'm thinking about future-proofing us a little here: how about we classify the `in_string`? 
I mean classify it as a local regular path, GCS path, or Azure path, or not a path at all.

src/cromshell/list_outputs/command.py
```diff
@@ -0,0 +1,200 @@
...
+        timeout=config.requests_connect_timeout,
+        verify=config.requests_verify_certs,
+        headers=http_utils.generate_headers(config),
+    )
+
+    if requests_out.ok:
+        return requests_out.json()
+    else:
+        http_utils.check_http_request_status_code(
+            short_error_message="Failed to retrieve workflow outputs.",
```
SHuang-Broad, line 80: It'd be better if we also include the ID of this workflow that fails.

src/cromshell/list_outputs/command.py
```diff
@@ -0,0 +1,200 @@
...
+            workflow_id=workflow_id, cromshell_config=config
+        )
+
+        if not detailed:
+            if json_summary:
+                io_utils.pretty_print_json(
+                    format_json=get_workflow_level_outputs(config).get("outputs")
+                )
+            else:
+                print_file_like_value_in_dict(
```
SHuang-Broad, line 49: Does this handle `Array[File]` outputs?

src/cromshell/list_outputs/command.py
```diff
@@ -0,0 +1,200 @@
...
+        if "subWorkflowMetadata" in calls_metadata[call][0]:
+            output_metadata[call] = []
+            for scatter in calls_metadata[call]:
+                output_metadata[call].append(
+                    filer_outputs_from_workflow_metadata(scatter["subWorkflowMetadata"])
+                )
+        else:
+            output_metadata[call] = []
+            for index in index_list:
+                output_metadata[call].append(index.get(extract_task_key))
```
SHuang-Broad, line 132: what if a task has no output? In theory it is possible. Then the `get` call will return `None`. Does the downstream handle `None` correctly? I think it's better if we handle that here.

SHuang-Broad (changes requested):
Let's also make sure #249 is reflected here.

tests/unit/test_list_outputs.py
bshifaw: oops, that should have been deleted

tests/integration/utility_test_functions.py
bshifaw: done, also changed name to subcommand_options

src/cromshell/list_outputs/command.py
bshifaw: It should handle `output_name: File` and `output_name: [File, File,..]`

src/cromshell/list_outputs/command.py
```diff
@@ -0,0 +1,200 @@
...
+    LOGGER.info("list-outputs")
+
+    return_code = 0
+
+    for workflow_id in workflow_ids:
+        command_setup_utils.resolve_workflow_id_and_server(
+            workflow_id=workflow_id, cromshell_config=config
+        )
+
+        if not detailed:
```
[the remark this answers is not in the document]
bshifaw: done

src/cromshell/list_outputs/command.py
bshifaw: good catch, done

src/cromshell/list_outputs/command.py
bshifaw: done

src/cromshell/utilities/command_setup_utils.py
```diff
@@ -17,5 +17,17 @@ def resolve_workflow_id_and_server(workflow_id: str, cromshell_config) -> str:
     http_utils.set_and_check_cromwell_server(
         config=cromshell_config, workflow_id=resolved_workflow_id
     )
+    set_workflow_id(workflow_id=resolved_workflow_id, cromshell_config=cromshell_config)
 
     return resolved_workflow_id
+
+
+def set_workflow_id(workflow_id: str, cromshell_config) -> None:
```
SHuang-Broad, line 25: I'm concerned about this.
What about commands that support multiple workflow IDs?

SHuang-Broad: resolved during code review meeting.

PR #249, opened 2023-03-17
Add option to not color json output to main command

bshifaw:
- Centralized json coloring setting to cromshellconfig.py, 
- Added option to disable coloring in main command
- By default json coloring is true
- JSON coloring is disabled if the output is pipped to file

### DIFF
--- a/src/cromshell/counts/command.py
+++ b/src/cromshell/counts/command.py
@@ -197,11 +197,12 @@ def print_call_status(call: str, indent: str, workflow_calls_metadata: dict) -> 
         )
 
 
-def print_task_status_summary(workflow_metadata: dict) -> None:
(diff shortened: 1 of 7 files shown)


src/cromshell/counts/command.py
```diff
@@ -197,11 +197,14 @@ def print_call_status(call: str, indent: str, workflow_calls_metadata: dict) ->
         )
 
 
-def print_task_status_summary(workflow_metadata: dict) -> None:
+def print_task_status_summary(
```
bshifaw, line 200: Added print_color parameter here mainly to help unit testing, allows me to disable coloring in testing so that its easier to evaluate the contents of the function output without dealing with the color formatting

SHuang-Broad (approved):
Good to go. Thanks for fixing this so quickly!

PR #256, opened 2023-05-01
Version Command fix

bshifaw:
The version command currently only displays the version of cromshell through Logger.info, but the logger by default is set to show only warnings/errors. 

This fix is to have the version command `print` the version and add unit/integration test for the command.


lbergelson (approved):
Good call.  Printing is the right move when someone is explicitly asking for the info.

PR #262, opened 2023-07-06
Fix out of date (un)installation commands

lbergelson:
Fix the installation commands that we forgot to update when we merged 2.0.0


SHuang-Broad (approved):
👍

PR #265, opened 2023-07-20
miniwdl validate command

bshifaw:
Added a command that validates wdl workflows using miniwdl package

### DIFF
--- a/src/cromshell/__main__.py
+++ b/src/cromshell/__main__.py
@@ -19,6 +19,7 @@ from .status import command as status
 from .submit import command as submit
 from .timing import command as timing
 from .update_server import command as update_server
+from .validate import command as validate
 
 # Version number is automatically set via bumpversion.
 # DO NOT MODIFY:
@@ -168,15 +169,16 @@ main_entry.add_command(abort.main)
 main_entry.add_command(alias.main)
 main_entry.add_command(cost.main)
 main_entry.add_command(counts.main)
-main_entry.add_command(status.main)
-main_entry.add_command(logs.main)
-main_entry.add_command(submit.main)
-main_entry.add_command(slim_metadata.main)
-main_entry.add_command(metadata.main)
-main_entry.add_command(update_server.main)
-main_entry.add_command(timing.main)
 main_entry.add_command(list.main)
 main_entry.add_command(list_outputs.main)
+main_entry.add_command(logs.main)
+main_entry.add_command(metadata.main)
+main_entry.add_command(slim_metadata.main)
+main_entry.add_command(status.main)
+main_entry.add_command(submit.main)
+main_entry.add_command(timing.main)
+main_entry.add_command(update_server.main)
+main_entry.add_command(validate.main)
 
 
 if __name__ == "__main__":
--- a/src/cromshell/validate/command.py
+++ b/src/cromshell/validate/command.py
@@ -1,0 +1,80 @@
+import logging
+from pathlib import Path
+
+import click
+from WDL import CLI as miniwdlCLI
+from WDL import Error as miniwdlError
+
+LOGGER = logging.getLogger(__name__)
+
+class ValidationFailedError(Exception):
+    pass
+
+@click.command(name="validate")
+@click.argument("wdl", type=click.Path(exists=True), required=True)
+@click.option(
+    "-d",
+    "--dependencies",
+    required=False,
+    multiple=True,
+    type=click.Path(exists=True),
+    help="Directory containing workflow source files that are "
+    "used to resolve local imports. (can supply multiple times)",
+)
+@click.option(
+    "-s",
+    "--strict",
+    is_flag=True,
+    default=False,
+    help="Exit with nonzero status code if any lint warnings are shown "
+         "(in addition to syntax and type errors)"
+)
+@click.option(
+    "-sup",
+    "--suppress",
+    required=False,
+    multiple=True,
+    help="Warnings to disable e.g. StringCoercion,NonemptyCoercion. (can supply multiple times)",
+)
+@click.pass_obj
+def main(config, wdl: Path, dependencies: tuple, strict: bool, suppress: tuple):
+    """
+    Validate a WDL workflow using miniwdl.
+    """
+
+    LOGGER.info("validate")
+
+    return_code = 0
+
+    # Check dependencies path exists:
+    for dep in dependencies:
+        Path(dep).resolve(strict=True)
+
+    # Validate the WDL file:
+    try:
+        miniwdlCLI.check(
+            uri=[str(wdl)],
+            path=list(dependencies),
+            strict=strict,
+            suppress=",".join([str(item) for item in suppress]) # Turn into string
+        )
+    except (
+            miniwdlError.SyntaxError,
+            miniwdlError.ImportError,
+            miniwdlError.ValidationError,
+            miniwdlError.MultipleValidationErrors,
+    ) as exn:
+        miniwdlCLI.print_error(exn)
+
+        if logging.getLogger().level == logging.DEBUG:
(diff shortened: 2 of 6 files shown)


tests/workflows/helloWorld.wdl
```diff
@@ -1,3 +1,5 @@
+version 1.0
```
bshifaw, line 1: Updated wdl to verison 1.0

tests/workflows/helloWorld.wdl
```diff
@@ -40,17 +44,16 @@ task HelloWorldTask {
...
-     String docker
-     Int? mem
-     Int? preemptible_attempts
-     Int? disk_space_gb
-     Int? cpu
-     Int? boot_disk_size_gb
-
+        String docker
+    # Runtime Options:
+        Int mem
```
bshifaw, line 51: miniwdl failed this workflow because it didn't like operating on optional int with an int
`mem * 1024`

lbergelson:
I would consider running both validations by default.  And doing both pre-submit.

src/cromshell/utilities/miniwdl_utils.py
```diff
@@ -0,0 +1,72 @@
...
+        suppress = []
+
+    for dep in dependencies:
+        Path(dep).resolve(strict=True)
+
+    try:
+        f = io.StringIO()
+        with redirect_stdout(f):
+            # Importing miniwdl.check here to help with stdout redirection
+            from WDL.CLI import check as miniwdl_check
```
bshifaw, line 46: Not a fan of importing in the middle of the script but this seemed the only way the stdout redirection would work. If not imported here miniwdl's listing function will print out several warnings. Which isn't helpful when running the function in the submit command

src/cromshell/utilities/miniwdl_utils.py
```diff
@@ -0,0 +1,72 @@
...
+
+            miniwdl_check(
+                uri=[str(wdl)],
+                path=list(dependencies),
+                strict=strict,
+                suppress=",".join([str(item) for item in suppress]),  # Turn into string
+            )
+
+        captured_outputs = f.getvalue()
+    except Exception as exn:
```
bshifaw, line 56: Here i wanted to capture any exception from miniwdl, mainly to reprint it in a cleaner way.

src/cromshell/utilities/miniwdl_utils.py
```diff
@@ -0,0 +1,72 @@
...
+    for dep in dependencies:
+        Path(dep).resolve(strict=True)
+
+    try:
+        f = io.StringIO()
+        with redirect_stdout(f):
+            # Importing miniwdl.check here to help with stdout redirection
+            from WDL.CLI import check as miniwdl_check
+
+            miniwdl_check(
```
lbergelson, line 48: I'm assuming miniwdl doesn't offer a clean way to get the results without just scraping the output logs?  It would be cleaner if we didn't have to do that.

src/cromshell/validate/command.py
```diff
@@ -0,0 +1,102 @@
...
+
+
+class MissingArgumentError(Exception):
+    pass
+
+
+@click.command(name="validate")
+@click.argument("wdl", type=click.Path(exists=True), required=True)
+@click.argument("wdl_json", type=click.Path(exists=True), required=False)
+@click.option(
```
lbergelson, line 23: In submit this is `"--dependencies-zip`. We should probably make this match.  (Or maybe we should change it in submit?)

lbergelson (changes requested):
@bshifaw I have a few comments but otherwise this looks great.  Thank you.

lbergelson:
My comments are all basically questions, only real actionable thing is changing the argument name.

src/cromshell/utilities/miniwdl_utils.py
```diff
@@ -0,0 +1,72 @@
...
+                uri=[str(wdl)],
+                path=list(dependencies),
+                strict=strict,
+                suppress=",".join([str(item) for item in suppress]),  # Turn into string
+            )
+
+        captured_outputs = f.getvalue()
+    except Exception as exn:
+        # Importing miniwdl.print_error here to help with stdout redirection
+        from WDL.CLI import print_error as miniwdl_print_error
```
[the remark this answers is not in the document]
bshifaw: yes

src/cromshell/validate/command.py
bshifaw: Miniwdl allows users to provide more than one folder to look for dependencies. 
I can make it match the `submit`command so that users can only provide one folder (or zip) to send to miniwdl.

src/cromshell/__main__.py
```diff
@@ -168,15 +169,16 @@ def print_version():
 main_entry.add_command(alias.main)
```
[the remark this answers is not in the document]
bshifaw: done

src/cromshell/utilities/miniwdl_utils.py
bshifaw: Not that I know of, having it directed to a variable didn't do anything

src/cromshell/validate/command.py
bshifaw: done, so now you can provide a zip or a path to a directory and it will provide it to miniwdl. I noticed that it doesn't handle nested workflows well (when the main workflow is in a child directory separate from its imported workflows). However it should be possible to reuse the same logic in the nested wdl PR to resolve it.

src/cromshell/validate/command.py
```diff
@@ -22,12 +22,12 @@ class MissingArgumentError(Exception):
...
 @click.option(
     "-d",
-    "--dependencies",
+    "--dependencies-zip",
     required=False,
-    multiple=True,
     type=click.Path(exists=True),
-    help="MiniWDL option: Directory containing workflow source files that are "
-    "used to resolve local imports. (can supply multiple times)",
+    help="MiniWDL option: ZIP file or directory containing workflow source files "
```
lbergelson, line 28: accessible to... what?  WHAT is it accessible to?  The mystery is killing me! :)

src/cromshell/validate/command.py
```diff
@@ -89,9 +89,11 @@ def main(
             wdl=str(wdl), wdl_json=str(wdl_json), config=config
         )
 
+    # Todo: Have it support nested directories after PR 268 is merged
+    # https://github.com/broadinstitute/cromshell/pull/268
```
lbergelson, line 93: 👍

tests/unit/conftest.py
```diff
@@ -33,3 +34,11 @@ def tmp_submission_file(mock_workflow_database_tsv, tmp_path):
     shutil.copyfile(mock_workflow_database_tsv, tmp_submission_file)
 
     return tmp_submission_file
+
+
+# Define a temporary directory for testing
+@pytest.fixture(scope="function")
+def temp_dir():
```
lbergelson, line 41: pytest fixtures are neat.

lbergelson (approved):
Looks good to me.

src/cromshell/validate/command.py
```diff
@@ -22,12 +22,12 @@ class MissingArgumentError(Exception):
...
-    "--dependencies",
+    "--dependencies-zip",
     required=False,
-    multiple=True,
     type=click.Path(exists=True),
-    help="MiniWDL option: Directory containing workflow source files that are "
-    "used to resolve local imports. (can supply multiple times)",
+    help="MiniWDL option: ZIP file or directory containing workflow source files "
+    "that are used to resolve local imports. This zip bundle will be "
+    "unpacked in a sandbox accessible to",
```
bshifaw, line 30:
```suggestion
    "unpacked in a sandbox accessible to this workflow.",
```

PR #268, opened 2023-07-26
Detect and flatten nested WDL directories

kvg:
One common problem we encounter is path resolution for WDLs within a nested directory structure. While a nested file organization is better for readability and maintainability, Cromwell cannot always resolve the file locations properly. Within Terra, dependencies are resolved via Dockstore. Outside of Terra, no convenient mechanism exists.

This PR adds such functionality to the `submit` command. First, we automatically detect when a workflow has relative imports (simply looking for the '../' indicator in the import path). Next, we create a temporary directory. We then recurse the dependency structure of the submitted WDL.  We rewrite each WDL to the temporary directory, adjusting the imports to point to the temporary directory as we go. We also ensure that the aliasing of WDLs remains correct by adding it where necessary.

To see this in action, consider the following directory structure. The `HelloWorkflow.wdl` imports files from directories above it, which works in Dockstore/Terra, but fails to run through Cromwell via cromshell.

```
[ 192]  /Users/kiran/repositories/aou-lr/wdl/
├── [ 352]  lib
│   ├── [ 480]  Utility
│   │   ├── [ 85K]  Utils.wdl
├── [  96]  pipelines
│   └── [ 580]  HelloWorkflow.wdl
├── [  96]  structs
│   └── [ 251]  Structs.wdl
└── [  96]  tasks
    └── [ 462]  HelloTask.wdl
```

To fix this, `cromshell submit` flattens out the directory structure within a temporary directory, as below.

```
[ 192]  /var/folders/jp/l0z21gnj4f531jw12fvm0bx80000gq/T/cromshell_alce64p_
├── [ 85K]  Users-kiran-repositories-aou-lr-wdl-lib-Utility-Utils.wdl
├── [ 655]  Users-kiran-repositories-aou-lr-wdl-pipelines-HelloWorkflow.wdl
├── [ 251]  Users-kiran-repositories-aou-lr-wdl-structs-Structs.wdl
└── [ 462]  Users-kiran-repositories-aou-lr-wdl-tasks-HelloTask.wdl
```

The submitted workflow, HelloWorkflow.wdl, is also rewritten:

```
version 1.0

import "Users-kiran-repositories-aou-lr-wdl-tasks-HelloTask.wdl" as Hello
import "Users-kiran-repositories-aou-lr-wdl-lib-Utility-Utils.wdl" as Utils

workflow HelloWorkflow {
    meta {
        description: "Example workflow."
    }

    parameter_meta {
        greeting: "The message to print"
    }

    input {
        String greeting
    }

    # Run a task locally defined in this repo
    call Hello.Print { input: message = greeting }

    # Run a task remotely defined in the long-read-pipelines repo
    call Utils.Sum { input: ints = [1, 2, 3] }

    output {
        String message = Print.text
        Int sum = Sum.sum
    }
}
```

And finally, the `dependencies_zip` parameter is overwritten to point to the temporary directory, enabling the automatic zipping of imports.

### DIFF
--- a/src/cromshell/submit/command.py
+++ b/src/cromshell/submit/command.py
@@ -4,6 +4,9 @@ import json
 import logging
 from datetime import datetime
 from pathlib import Path, PurePath
+import tempfile
+
+import os
 
 import click
 import requests
@@ -64,6 +67,11 @@ def main(config, wdl, wdl_json, options_json, dependencies_zip, no_validation):
     LOGGER.info("submit")
 
     http_utils.assert_can_communicate_with_server(config=config)
+
+    if io_utils.has_nested_dependencies(wdl):
+        tempdir = tempfile.TemporaryDirectory(prefix='cromshell_')
+        wdl = io_utils.flatten_nested_dependencies(tempdir, wdl)
+        dependencies_zip = tempdir.name
 
     if no_validation:
         LOGGER.info("Skipping WDL validation")
--- a/src/cromshell/utilities/io_utils.py
+++ b/src/cromshell/utilities/io_utils.py
@@ -1,11 +1,13 @@
 import json
 import logging
 import re
+import os
 import shutil
 from contextlib import nullcontext
 from io import BytesIO
 from pathlib import Path
-from typing import BinaryIO, List, Union
+import tempfile
+from typing import BinaryIO, List, Union, Tuple
 from zipfile import ZIP_DEFLATED, ZipFile
 
 from pygments import formatters, highlight, lexers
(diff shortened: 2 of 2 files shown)


src/cromshell/submit/command.py
```diff
@@ -65,6 +68,11 @@ def main(config, wdl, wdl_json, options_json, dependencies_zip, no_validation):
 
     http_utils.assert_can_communicate_with_server(config=config)
 
+    if io_utils.has_nested_dependencies(wdl):
```
bshifaw, line 71: It would be a good idea to have the option to disable this feature.
Also Logger.debug message saying this is taking place within the if statement

src/cromshell/utilities/io_utils.py
```diff
@@ -1,11 +1,13 @@
 import json
 import logging
 import re
+import os
```
bshifaw, line 4: If possible use avoid using os package and use pathlib.Path library for file/directly handling. The pathlib module is a newer module in Python that provides a more object-oriented way of handling file paths. This makes it more intuitive and easier to use, and it also provides a number of features that are not available in the os module.

src/cromshell/utilities/io_utils.py
```diff
@@ -101,6 +103,59 @@ def assert_path_is_not_empty(path: Union[str, Path], description: str) -> None:
...
 
+def has_nested_dependencies(wdl_path: str or Path) -> bool:
+    """Determine if a WDL has any nested imports."""
+
+    with open(wdl_path, 'r') as rf:
+        for l in rf:
+            if l.startswith('import'):
+                m = re.match(r'import "(.+)"', l)
+
+                if "../" in m.group(1):
```
bshifaw, line 114: An edge case that this function would miss would be nested wdl that uses absolute paths for their imports. Highly unlikely though as I haven't seen wdls like that often or at all.

src/cromshell/utilities/io_utils.py
```diff
@@ -101,6 +103,59 @@ def assert_path_is_not_empty(path: Union[str, Path], description: str) -> None:
...
+                    return True
+
+    return False
+
+
+def get_flattened_filename(tempdir: tempfile.TemporaryDirectory, wdl_path: str or Path):
```
bshifaw, line 120: add the output type being returned by the function

src/cromshell/utilities/io_utils.py
```diff
@@ -101,6 +103,59 @@ def assert_path_is_not_empty(path: Union[str, Path], description: str) -> None:
...
+    """Generate the filename to use for flattened WDL files."""
```
bshifaw, line 121: double-check my description 
```suggestion
    """ Generate hyphen-separated path to use for flattened WDL file path.
    For example: 
    tempdir: /path/2/tempdir/ and wdl_path: /dir/path/2/wdl.wdl
    returns: /path/2/tempdir/dir-path-2-wdl.wdl
    """
```

src/cromshell/utilities/io_utils.py
```diff
@@ -101,6 +103,59 @@ def assert_path_is_not_empty(path: Union[str, Path], description: str) -> None:
...
+
+    wdl_dir = os.path.dirname(wdl_path)
+
+    new_wdl_path = get_flattened_filename(tempdir, wdl_path)
+
+    with open(wdl_path, 'r') as rf, open(new_wdl_path, 'w') as wf:
+        for l in rf:
+            if l.startswith('import'):
+                m = re.match(r'import "(.+)"', l)
+
```
bshifaw, line 143: Not needed but to improve readability it would be helpful to put m.group(1) into a variable and use it in the lines below. 
```suggestion
                imported_wdl_name = m.group(1)
```

bshifaw (changes requested):
- Added some suggestions.
- If viable unit tests for the new functions and an integration tests would be valuable. For the integration tests I think you can simply add a simple nested wdl to the workflows `/tests/workflows` dir and add the workflow to the exists tests in the paramterazied input for [test_submit](https://github.com/broadinstitute/cromshell/blob/6e0ac92ffebe90368e738da6b5063dc3f1d39e49/tests/integration/test_submit.py#L32)
- Also to automatically fix linting issues you can use `tox -e lint-edit`

src/cromshell/utilities/io_utils.py
```diff
@@ -1,7 +1,10 @@
 import json
 import logging
 import re
+
```
bshifaw, line 4:
```suggestion
```

src/cromshell/utilities/io_utils.py
```diff
@@ -1,7 +1,10 @@
...
+# import os
```
bshifaw, line 5:
```suggestion
```

Issue #277, opened 2024-01-11
Local pip install fails


sjfleming:
If I run
```console
git clone git@github.com:broadinstitute/cromshell.git
cd cromshell
pip install -e .
```
it fails to install.

I see this
```console
Obtaining file:///Users/sfleming/Documents/Github/cromshell
  Installing build dependencies ... done
  Checking if build backend supports build_editable ... done
  Getting requirements to build editable ... error
  error: subprocess-exited-with-error
  
  × Getting requirements to build editable did not run successfully.
  │ exit code: 1
  ╰─> [173 lines of output]
      /private/var/folders/zd/jpvyhqzx6z19xsdgl7f9fjy40000gq/T/pip-build-env-p01ww7jv/overlay/lib/python3.10/site-packages/setuptools/config/_apply_pyprojecttoml.py:75: _MissingDynamic: `description` defined outside of `pyproject.toml` is ignored.
      !!
      
              ********************************************************************************
              The following seems to be defined outside of `pyproject.toml`:
      
              `description = 'Command Line Interface (CLI) for Cromwell servers'`
      
              According to the spec (see the link below), however, setuptools CANNOT
              consider this value unless `description` is listed as `dynamic`.
      
              https://packaging.python.org/en/latest/specifications/declaring-project-metadata/
      
              To prevent this problem, you can list `description` under `dynamic` or alternatively
              remove the `[project]` table from your file and rely entirely on other means of
              configuration.
              ********************************************************************************
      
      !!
        _handle_missing_dynamic(dist, project_table)
      /private/var/folders/zd/jpvyhqzx6z19xsdgl7f9fjy40000gq/T/pip-build-env-p01ww7jv/overlay/lib/python3.10/site-packages/setuptools/config/_apply_pyprojecttoml.py:75: _MissingDynamic: `requires-python` defined outside of `pyproject.toml` is ignored.
      !!
      
              ********************************************************************************
              The following seems to be defined outside of `pyproject.toml`:
      
              `requires-python = '>=3.7'`
      
              According to the spec (see the link below), however, setuptools CANNOT
              consider this value unless `requires-python` is listed as `dynamic`.
      
              https://packaging.python.org/en/latest/specifications/declaring-project-metadata/
      
              To prevent this problem, you can list `requires-python` under `dynamic` or alternatively
              remove the `[project]` table from your file and rely entirely on other means of
              configuration.
              ********************************************************************************
      
      !!
        _handle_missing_dynamic(dist, project_table)
[...]

PR #278, opened 2024-01-19
Fix cost (for methods cromwell), add explicit example

sjfleming:
I didn't have a clue how to get `cromshell cost` to work.  @SHuang-Broad helped me out.

I added the information I needed to the README.md so others can get it to work.

There also seemed to be a bug in `cromshell cost`: the code expected a table column called `partition_time` but the table here (Methods cromwell server v77 location)
https://console.cloud.google.com/bigquery?project=broad-dsde-methods&organizationId=548622027621&j=bq:US:bquxjob_85a71e8_16fa747f47d&page=queryresults&ws=!1m10!1m4!4m3!1sbroad-dsde-methods!2sMethods_billing_dump!3sgcp_billing_export_v1_009C7D_923007_219A6F!1m4!1m3!1sbroad-dsde-methods!2sbquxjob_85a71e8_16fa747f47d!3sUS
has a column called `export_time` instead.

When I changed the code from `partition_time` to `export_time`, I was able to get the command to work and see my job costs.


sjfleming:
The test errors I'm getting appear to be #277 

The github actions seem to be failing when installing the repo, unless I'm misunderstanding the logs.

README.md
```diff
@@ -110,6 +110,15 @@ Cromshell is a CLI for submitting workflows to a Cromwell server and monitoring/
      * Requires the `bq_cost_table` key to exist in the cromshell 
        configuration file and have a value equal to the BigQuery cost table 
        for your GCP billing project.
+       * For example, using the Methods cromwell server, your `~/.cromshell/cromshell_config.json` 
+         should be as follows:
+         ```json
+         {
+           "cromwell_server": "https://cromwell-v77.dsde-methods.broadinstitute.org/",
+           "requests_timeout": 5,
+           "bq_cost_table": "Methods_billing_dump.gcp_billing_export_v1_009C7D_923007_219A6F"
```
SHuang-Broad, line 119: I'd prefer we hide these values from the open internet, just in case...

SHuang-Broad: so how about the following:
`"<bq_dataset_hosting_the_billing_datatable>.<the_billing_datatable>"`

src/cromshell/cost/command.py
```diff
@@ -182,15 +182,15 @@ def create_bq_query(detailed: bool, bq_cost_table: str) -> str:
                 AND task.key LIKE "wdl-task-name"
                 AND wfid.key LIKE "cromwell-workflow-id"
                 AND wfid.value like @workflow_id
-                AND partition_time BETWEEN @start_date AND @end_date
+                AND export_time BETWEEN @start_date AND @end_date
```
SHuang-Broad, line 185: I looked at our cost tabe and do see this key in it, instead of the `partition_time` key. But I wonder why it was `partition_time` at the beginning, starting from the shell version. Did Google change things underneath us?

`export_time `is documented here, but not `partition_time`
https://cloud.google.com/billing/docs/how-to/export-data-bigquery-tables/detailed-usage

And confusing the situation, even more, are the two other `*time` values: `usage_start_time` and `usage_end_time`. And now I wonder, should we use them instead of `export_time`?

README.md
sjfleming: We could do that if we need to yeah

sjfleming: (But if there's no risk in exposing these, I'd like the example to be 100% explicit.  Because I never would have foudn those values.
  So... is there a risk?)

sjfleming: Or... better... if we included directions for _how_ to find that table.

sjfleming: But I don't know what those directions are

bshifaw:
The unit/integration tests are failing for an unrelated reason and is being fixed [here](https://github.com/broadinstitute/cromshell/pull/279). Once that PR is merged, updating your branch from main should pass your tests

bshifaw (changes requested):
I've pulled in the changes to fix the previous git action test fails.
Now, just need to update the unit test script that checks the for the expected query. Update it the "export_time" 
https://github.com/broadinstitute/cromshell/blob/f1344e208a8a0a1375651f70ce8e0261ff6747b9/tests/unit/test_cost.py#L26C21-L26C35
https://github.com/broadinstitute/cromshell/blob/f1344e208a8a0a1375651f70ce8e0261ff6747b9/tests/unit/test_cost.py#L37

sjfleming:
I think it's ready pending @SHuang-Broad 's comment above about hiding the values from the README.  Not sure what the consensus is there.  I see Steve's point, but also if the example is less explicit, I would never be able to figure out how to use `cost`, because I don't know how to find the location of the bq table.

SHuang-Broad:
@sjfleming 
How about this:

```
The URL for accessing the BigQuery cost table is formatted as 
https://console.cloud.google.com/bigquery?project=<your_gcp_billing_project>
```
where you can see the table like displayed in the following figure (actual project hidden)
Screenshot 2024-03-29 at 2 29 18 PM

Click on the table itself (in this case, `gcp_billing_...`) and open the "DETAILS" tab, you'll find the exact path to the table in the "Table ID" section.
Screenshot 2024-03-29 at 2 33 59 PM

sjfleming:
Thanks @SHuang-Broad !

Alright how about this README?  (Tried to just use one picture, but I think these are super helpful.)

SHuang-Broad:
Looks good to me! Thanks for getting all these together!

bshifaw (approved):

PR #279, opened 2024-02-22
Replacing Setup.py with Pyproject.py

bshifaw:
The python community is moving towards pyproject as the standard method of specifying a project metadata.
> Starting with [PEP 621](https://peps.python.org/pep-0621/), the Python community selected pyproject.toml as a standard way of specifying project metadata
[link](https://setuptools.pypa.io/en/latest/userguide/pyproject_config.html)

The Cromshell repo has been using a mix of pyproject.py and setup.py, this is currently causing issue during installation.

```
integration: packaging backend failed (code=1), with AttributeError: 'NoneType' object has no attribute 'split'
/home/runner/work/cromshell/cromshell/.tox/.pkg/lib/python3.8/site-packages/setuptools/config/_apply_pyprojecttoml.py:75: _MissingDynamic: `description` defined outside of `pyproject.toml` is ignored.
!!

        ********************************************************************************
        The following seems to be defined outside of `pyproject.toml`:

        `description = 'Command Line Interface (CLI) for Cromwell servers'`

        According to the spec (see the link below), however, setuptools CANNOT
        consider this value unless `description` is listed as `dynamic`.

        https://packaging.python.org/en/latest/specifications/declaring-project-metadata/

        To prevent this problem, you can list `description` under `dynamic` or alternatively
        remove the `[project]` table from your file and rely entirely on other means of
        configuration.
        ********************************************************************************
```

Changes in this repo replaceses setup.py completely with pyproject.toml, fixing the installation errors.


SHuang-Broad (commented):
Tests seem to be failing.

.github/workflows/integration_tests.yml
```diff
@@ -9,7 +9,7 @@ jobs:
     runs-on: ubuntu-latest
     strategy:
       matrix:
-        python: [ '3.7', '3.8', '3.9', '3.10' ]
```
bshifaw, line 12: 3.7 is no longer being supported and support for 3.8 is ending this year: https://devguide.python.org/versions/

dev-requirements.txt
```diff
@@ -1,13 +1,13 @@
-isort==4.3.21
+isort==5.13.2
 tox
 wheel
 bumpversion
 twine
 black
 flake8
 pylint
+tomli==2.0.1
```
bshifaw, line 9: needed for tox validation of pyproject.toml

dev-requirements.txt
```diff
@@ -1,13 +1,13 @@
...
 cython
 sphinx
 sphinx-rtd-theme
-mypy==0.770
--r requirements.txt
```
bshifaw, line 13: pyproject.toml does not support recursive package installation

.github/workflows/pypi_packaging.yml
```diff
@@ -80,9 +80,6 @@ jobs:
     - name: Setup Env
       run: python3 -m pip install --upgrade pip build
 
-    - name: Validate setup.py
```
bshifaw, line 83: There is no integrated solution for validating the pyproject.toml file. There are some third-party application but they are unstable. 
I've noticed though that tox run excutting its test automatically does a basic level check of the toml file using not a cli command but via package called "tomli". This has been added to the dev-requirments.txt file

dev-requirements.txt
lbergelson: How do transitive dependencies get installed then?

pyproject.toml
```diff
@@ -1,3 +1,5 @@
+
+ignore_missing_imports = true
```
lbergelson, line 2: Can you explain this one to me?  Don't we need all our imports?

.github/workflows/integration_tests.yml
lbergelson: Can you no longer install on python 3.7 /8 now? Or is it just not being tested.

lbergelson: Just out of an abundance of caution we might want to check what version of python the terra and all of us images are using.

lbergelson (approved):
@bshifaw Looks fine to me, but I definitely don't understand the nuance here.

I think we should probably check with the people using it for the all of us workspaces to make sure they have 3.9 available.

bshifaw:
> @bshifaw Looks fine to me, but I definitely don't understand the nuance here.
> 
> I think we should probably check with the people using it for the all of us workspaces to make sure they have 3.9 available.

This update was to help fix the installation on AOU workspaces. When running their cromshell tutorial, the installation included a cell that updated the cromshell version by pip - git (pip install git+https://github.com/broadinstitute/cromshell). This caused the installation error mentioned in the PR description. Their python version is 3.10.3. 

Also, to confirm I just now installed cromshell in an AOU notebook using this specific branch; the instillation now completes successfully.

dev-requirements.txt
bshifaw: I mean to say pyproject.toml requires a dependency txt file to list packages directly, and not point to another dependency file.

Transitive dependencies are still handled by pip during installation. The requiremnts.txt file is listed as core dependencies in the pyprojec.toml,  and pip will install the transitive dependencies. The dev-requirements and test-requirements are listed as optional dependencies

pyproject.toml
bshifaw: Thanks for this catch, its suppose  to be under `[tool.mypy."pytest.*"]`
When ignore_missing_imports is set to true, mypy will not report an error when an imported module cannot be found. This is to replicate the setting that was originally found in the setup.cfg file

tox.ini
```diff
@@ -49,9 +53,9 @@ deps =
 		flake8
 
 commands =
-    black --check --diff --target-version py38 src tests setup.py
-    isort --check-only --profile black --diff -rc tests src setup.py
```
bshifaw, line 53: `-rc` for isort is deprecated because options now happen by default

.github/workflows/integration_tests.yml
bshifaw: Terra/AOU uses python3.10.12
cromshell on p3.7 is no longer being tested

I had some trouble installing dev packages on py3.11 because  mypy==0.770 was not supported. 
So I updated this to the latest [version](https://pypi.org/project/mypy/), the latest version doesn't work on 3.7. None of the packages in the requirements have been updated so cromshell should still work on py3.7 but if anyone is developing on 3.7 they'll run into package issues.

dev-requirements.txt
lbergelson: Thank you for the clarification!

